PR #33, opened 2015-01-03
Allow you to specify a serie when creating an event

simong:
Builds on #30

PR #34, opened 2015-01-05
Allow you to specify a serie when creating an event

simong:
Similar as #33 but without old commits


nicolaasmatthijs:
Looks good. Merged.

PR #35, opened 2015-01-05
Updating the organisers of an event

simong:


node_modules/gh-core/lib/db.js
```diff
@@ -180,9 +180,11 @@ var _setUpModel = function(sequelize) {
                 return true;
             },
             'toJSON': function() {
+                var json = _.clone(this.dataValues);
```
nicolaasmatthijs: Is there any particular reason why the cloning is required?

node_modules/gh-events/lib/api.js
```diff
@@ -300,6 +300,87 @@ var updateEvent = module.exports.updateEvent = function(ctx, id, update, callbac
...
+ */
+var updateEventOrganisers = module.exports.updateEventOrganisers = function(ctx, id, update, callback) {
+    update = update || {};
+
+    var validator = new Validator();
+    validator.check(null, {'code': 401, 'msg': 'Only authenticated users can update events'}).isLoggedInUser(ctx);
+    validator.check(id, {'code': 400, 'msg': 'A valid event id must be provided'}).isInt();
+    validator.check(_.keys(update).length, {'code': 400, 'msg': 'At least 1 update value must be provided'}).min(1);
+    _.each(update, function(val, user) {
+        update[user] = GrasshopperUtil.getBooleanParam(update[user]);
```
nicolaasmatthijs: Could just do `GrasshopperUtil.getBooleanParam(val)` here

node_modules/gh-events/lib/api.js
```diff
@@ -300,6 +300,87 @@ var updateEvent = module.exports.updateEvent = function(ctx, id, update, callbac
...
+    // Ensure the event exists
+    getEvent(ctx, id, function(err, event) {
+        if (err) {
+            return callback(err);
+        }
+
+        // Get the user ids, if any, from the update
+        var userIds = _.chain(update)
+            .keys()
+            .map(function(user) {
```
nicolaasmatthijs: We can probably do the map and filter in one go

node_modules/gh-events/lib/api.js
```diff
@@ -300,6 +300,87 @@ var updateEvent = module.exports.updateEvent = function(ctx, id, update, callbac
...
+    }
+
+    // Ensure the event exists
+    getEvent(ctx, id, function(err, event) {
+        if (err) {
+            return callback(err);
+        }
+
+        // Get the user ids, if any, from the update
+        var userIds = _.chain(update)
```
nicolaasmatthijs: Wouldn't it be cleaner to split off the actual user ids from the other organisers into separate variables as early on as possible to get cleaner method signatures?

node_modules/gh-events/lib/api.js
```diff
@@ -300,6 +300,87 @@ var updateEvent = module.exports.updateEvent = function(ctx, id, update, callbac
...
+
+            // Ensure that the current user is allowed to update the event
+            EventsAuthz.canUpdateEvent(ctx, event, function(err, canUpdateEvent) {
+                if (err) {
+                    return callback(err);
+                } else if (!canUpdateEvent) {
+                    return callback({'code': 401, 'msg': 'You are not allowed to update this event'});
+                }
+
+                // TODO: Add a "confirm"-flow for the organisers, so regular users can't add other
```
nicolaasmatthijs: I don't think we need this at the moment. We can add it when a production need arises

node_modules/gh-events/lib/api.js
```diff
@@ -300,6 +300,87 @@ var updateEvent = module.exports.updateEvent = function(ctx, id, update, callbac
...
+                // users willy-nilly to their own events?
+
+                // The DAO will make inline changes to the `event` instance. If we wish to provide
+                // the old instance in the emitted event we need to take a copy here. We cannot take
+                // a copy of the full `event` instance as that has things such as connections and/or
+                // transactions associated to it, which could lead to potentially confusing situations
+                // further down the future
+                var oldEvent = _.clone(event.toJSON());
```
nicolaasmatthijs: We're cloning it twice at the moment

node_modules/gh-events/lib/internal/dao.js
```diff
@@ -158,6 +158,81 @@ var updateEvent = module.exports.updateEvent = function(event, update, callback)
 };
 
 /**
+ * Update the organisers of an event
+ *
+ * @param  {Event}      event               The event to update
+ * @param  {Object}     update              The updates to persist. Keys that are numbers are considered to be references to users, anything else is considered to be a plain-text organiser
```
nicolaasmatthijs: I think it would be nicer to have them split off here (maybe even split them off in the REST API)

node_modules/gh-events/lib/internal/dao.js
```diff
@@ -158,6 +158,81 @@ var updateEvent = module.exports.updateEvent = function(event, update, callback)
...
+ * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error object, if any
+ * @param  {Event}      callback.event      The updated event
+ */
+var updateEventOrganisers = module.exports.updateEventOrganisers = function(event, update, callback) {
+    // TODO: Transactions
+    // Split out the update in organisers that are address by user id or provided as plain text
```
nicolaasmatthijs: typo

node_modules/gh-events/lib/restmodel.js
```diff
@@ -56,21 +56,6 @@
...
- * @Required  []
- * @Property  {boolean}                     {organiserUser}     Whether the named user should be an organiser
- */
-
-/**
- * @RESTModel EventOrganiserUpdateUser
- *
  * @Required  []
- * @Property  {boolean}                     {organiserOther}    Whether the named user should be an organiser
+ * @Property  {boolean}                     {organiserUser}     Whether the user should be an organiser
```
nicolaasmatthijs: Is there any reason why we're moving away from having 2 separate parameters for normal and other organisers? That feels like a cleaner REST API to me.

nicolaasmatthijs: Regardless, if we keep the current approach, this should provide detail about the id/string distinction

node_modules/gh-events/tests/test-organisers.js
```diff
@@ -0,0 +1,205 @@
...
+                    // Add Nico as an organiser
+                    var update = {};
+                    update[nico.profile.id] = true;
+                    EventsTestUtil.assertUpdateEventOrganisers(simon.client, event.id, update, function() {
+
+                        // Add a plain-text organiser
+                        update = {'Mr Doo Little': true};
+                        EventsTestUtil.assertUpdateEventOrganisers(simon.client, event.id, update, function() {
+
+                            // Add Sam and a plain-text organiser
```
nicolaasmatthijs: Documentation error

node_modules/gh-events/tests/test-organisers.js
```diff
@@ -0,0 +1,205 @@
...
+        });
+
+        /**
+         * Test that verifies that users from other applications cannot be added as organisers to an event
+         */
+        it('verify users from other applications cannot be added as organisers to an event', function(callback) {
+            TestsUtil.generateTestUsers(global.tests.apps.cam2013, 1, false, function(simon) {
+                TestsUtil.generateTestUsers(global.tests.apps.oxford2014, 2, false, function(nico) {
+
+                    // Create an event, that Nico organises
```
nicolaasmatthijs: Remove `,`

node_modules/gh-events/tests/util.js
```diff
@@ -227,6 +227,67 @@ var assertUpdateEventFails = module.exports.assertUpdateEventFails = function(cl
 };
 
 /**
+ * Assert that the organisers for an event can be updated
+ *
+ * @param  {RestClient}         client                          The REST client to make the request with
+ * @param  {Number}             id                              The id of the event whose organisers to update
```
nicolaasmatthijs: Should be consistent with the REST API: `The id of the event to update`

nicolaasmatthijs: Should be the case in the other places as well

nicolaasmatthijs:

node_modules/gh-events/lib/restmodel.js
simong: Because what was outlined in those docs isn't actually possible through REST. I think you'd have to submit a JSON blob and parse it on the backend in order to achieve that.

node_modules/gh-core/lib/db.js
simong: There are places outside of REST serialisation now where we do `instance.toJSON()`. Serialising an instance shouldn't have any side-effects such as properties being gone.

node_modules/gh-events/tests/test-organisers.js
simong: I don't see it :/

simong:
Followed-up, back to @nicolaasmatthijs

node_modules/gh-events/tests/test-organisers.js
nicolaasmatthijs: Oh dear. I had totally read `as a plain-text organiser`.

nicolaasmatthijs:
Left some minor remaining feedback. Back to @simong

simong:

nicolaasmatthijs:
Looks good. Merged.

PR #43, opened 2015-01-13
Added a test that a user cannot re-use another user’s email address when updating their profile

simong:
At some point we'll want to add in email verification though

### DIFF
--- a/node_modules/gh-users/tests/test-users.js
+++ b/node_modules/gh-users/tests/test-users.js
@@ -149,5 +149,18 @@ describe('Users', function() {
                 });
             });
         });
+
+        /**
+         * Test that verifies that a user cannot use another user's email address
+         */
+        it('verify a user cannot use another user\'s email address', function(callback) {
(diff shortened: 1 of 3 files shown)


node_modules/gh-users/tests/test-users.js
```diff
@@ -149,5 +149,18 @@ describe('Users', function() {
...
             });
         });
+
+        /**
+         * Test that verifies that a user cannot use another user's email address
+         */
+        it('verify a user cannot use another user\'s email address', function(callback) {
+            TestsUtil.generateTestUsers(global.tests.apps.cam2014, 2, false, function(simon, nico) {
+
+                // Nico shouldn't be able to re-use Simon's email address
```
nicolaasmatthijs: Can we also do a test where a new user is created using an existing email address?

nicolaasmatthijs:

node_modules/gh-users/tests/test-users.js
simong: Creating users is one of those areas that hasn't received much test love yet somehow ;-)

nicolaasmatthijs:
Looks good. Merged.

PR #45, opened 2015-01-15
Implemented getUser and hiding of personal information

simong:
Supports the REST.getUser logic. A `hide` function is added to each
`User` instance which will hide personal information from users who are
not allowed to see it.

### DIFF
--- a/node_modules/gh-core/lib/db.js
+++ b/node_modules/gh-core/lib/db.js
@@ -253,6 +253,23 @@ var _setUpModel = function(sequelize) {
             'isGlobalAdmin': function() {
                 return false;
             },
+            'hide': function(ctx) {
+                // There's no need to hide information if the current user is requesting their
+                // own profile or the current user is can administrate the user's application
+                if (ctx.user && (ctx.user.id === this.id || ctx.user.canAdmin(this.AppId))) {
+                    return;
+                }
+
+                // Otherwise we need to hide some information
+                delete this.dataValues.authenticationStrategy;
+                delete this.dataValues.password;
+                delete this.dataValues.calendarToken;
+                delete this.dataValues.email;
+                delete this.dataValues.emailPreference;
+                delete this.dataValues.isAdmin;
+                delete this.dataValues.shibbolethId;
+                delete this.dataValues.termsAndConditions;
+            },
             'toJSON': function() {
                 var json = _.clone(this.dataValues);
 
--- a/node_modules/gh-core/lib/db.js
+++ b/node_modules/gh-core/lib/db.js
@@ -253,9 +253,9 @@ var _setUpModel = function(sequelize) {
             'isGlobalAdmin': function() {
                 return false;
             },
-            'hide': function(ctx) {
+            'hidePersonalInformation': function(ctx) {
                 // There's no need to hide information if the current user is requesting their
-                // own profile or the current user is can administrate the user's application
+                // own profile or the current user can administrate the user's application
                 if (ctx.user && (ctx.user.id === this.id || ctx.user.canAdmin(this.AppId))) {
                     return;
                 }
--- a/node_modules/gh-events/lib/api.js
+++ b/node_modules/gh-events/lib/api.js
@@ -197,6 +197,11 @@ var getEvent = module.exports.getEvent = function(ctx, id, callback) {
             return callback({'code': 401, 'msg': 'This event cannot be retrieved on this application'});
         }
 
+        // Hide organisers their personal information
+        _.each(event.Organisers, function(organiser) {
+            organiser.hide(ctx);
+        });
+
         return callback(null, event);
     });
 };
--- a/node_modules/gh-events/lib/api.js
+++ b/node_modules/gh-events/lib/api.js
@@ -24,6 +24,7 @@ var GroupsAPI = require('gh-groups');
 var GroupsDAO = require('gh-groups/lib/internal/dao');
 var log = require('gh-core/lib/logger').logger('gh-events');
 var SeriesDAO = require('gh-series/lib/internal/dao');
+var UsersAPI = require('gh-users');
 var UsersDAO = require('gh-users/lib/internal/dao');
 var Validator = require('gh-core/lib/validator').Validator;
 
@@ -198,9 +199,7 @@ var getEvent = module.exports.getEvent = function(ctx, id, callback) {
         }
 
         // Hide organisers their personal information
-        _.each(event.Organisers, function(organiser) {
-            organiser.hide(ctx);
-        });
+        UsersAPI.hideUsersPersonalInformation(ctx, event.Organisers);
 
         return callback(null, event);
     });
--- a/node_modules/gh-events/tests/test-organisers.js
+++ b/node_modules/gh-events/tests/test-organisers.js
@@ -22,6 +22,7 @@ var moment = require('moment');
 
 var GroupsTestUtil = require('gh-groups/tests/util');
 var TestsUtil = require('gh-tests');
+var UsersTestsUtil = require('gh-users/tests/util');
 
 var EventsTestUtil = require('./util');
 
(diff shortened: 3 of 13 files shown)


node_modules/gh-core/lib/db.js
```diff
@@ -253,6 +253,23 @@ var _setUpModel = function(sequelize) {
             'isGlobalAdmin': function() {
                 return false;
             },
+            'hide': function(ctx) {
+                // There's no need to hide information if the current user is requesting their
+                // own profile or the current user is can administrate the user's application
```
nicolaasmatthijs: typo

node_modules/gh-core/lib/db.js
```diff
@@ -253,6 +253,23 @@ var _setUpModel = function(sequelize) {
             'isGlobalAdmin': function() {
                 return false;
             },
+            'hide': function(ctx) {
```
nicolaasmatthijs: `hide` isn't really accurate. Something like `publicProfile` or `profile` might be better?

node_modules/gh-core/lib/db.js
```diff
@@ -253,6 +253,23 @@ var _setUpModel = function(sequelize) {
...
+                // There's no need to hide information if the current user is requesting their
+                // own profile or the current user is can administrate the user's application
+                if (ctx.user && (ctx.user.id === this.id || ctx.user.canAdmin(this.AppId))) {
+                    return;
+                }
+
+                // Otherwise we need to hide some information
+                delete this.dataValues.authenticationStrategy;
```
nicolaasmatthijs: Would it be safer to whitelist instead of blacklist?

node_modules/gh-events/lib/api.js
```diff
@@ -197,6 +197,11 @@ var getEvent = module.exports.getEvent = function(ctx, id, callback) {
             return callback({'code': 401, 'msg': 'This event cannot be retrieved on this application'});
         }
 
+        // Hide organisers their personal information
+        _.each(event.Organisers, function(organiser) {
```
nicolaasmatthijs: Wouldn't it be better to expose a `hideUsers` function in the users API and call that everywhere?

node_modules/gh-events/tests/test-organisers.js
```diff
@@ -29,6 +30,46 @@ describe('Events', function() {
...
+                    update[nico.profile.id] = true;
+                    EventsTestUtil.assertUpdateEventOrganisers(simon.client, event.id, update, function() {
+
+                        // Get the event
+                        EventsTestUtil.assertGetEvent(simon.client, event.id, null, function(event) {
+                            assert.strictEqual(event.organisers.length, 2);
+
+                            // Verify Nico's information is hidden
+                            var organisersById = _.indexBy(event.organisers, 'id');
+                            assert.strictEqual(organisersById[nico.profile.id].displayName, nico.profile.displayName);
```
nicolaasmatthijs: Maybe we can create a `assertPublicUser` function in the user tests util and call that to verify that the appropriate properties are present

node_modules/gh-series/tests/test-events.js
```diff
@@ -494,5 +494,47 @@ describe('Event series', function() {
...
+                    var update = {};
+                    update[nico.profile.id] = true;
+                    EventsTestUtil.assertUpdateEventOrganisers(simon.client, event.id, update, function() {
+
+                        // Get the series events
+                        SeriesTestUtil.assertGetSeriesEvents(simon.client, serie.id, null, null, null, null, function(events) {
+
+                            // Assert nico's personal information is hidden
+                            var seenNico = false;
+                            _.each(events.results, function(event) {
```
nicolaasmatthijs: The `_.indexBy` approach from above seems simpler

node_modules/gh-users/lib/api.js
```diff
@@ -50,11 +50,24 @@ var init = module.exports.init = function(callback) {
  * @param  {Object}         callback.err                    An error object, if any
```
nicolaasmatthijs: Need to add `ctx` in JSDocs

node_modules/gh-users/lib/api.js
```diff
@@ -50,11 +50,24 @@ var init = module.exports.init = function(callback) {
...
  * @param  {User}           callback.user                   The requested user
  */
-var getUser = module.exports.getUser = function(id, callback) {
-    // TODO: id validation
-    // TODO: hide calendarToken
+var getUser = module.exports.getUser = function(ctx, id, callback) {
+    var validator = new Validator();
+    validator.check(null, {'code': 401, 'msg': 'You need to be authenticated in order to get a user'}).isLoggedInUser(ctx);
```
nicolaasmatthijs: Should be `Only authenticated users can get a user` to be consistent

node_modules/gh-users/lib/api.js
```diff
@@ -50,11 +50,24 @@ var init = module.exports.init = function(callback) {
...
+    validator.check(id, {'code': 400, 'msg': 'A valid user id must be provided'}).isInt();
+    if (validator.hasErrors()) {
+        return callback(validator.getFirstError());
+    }
+
+    UsersDAO.getUser(id, function(err, user) {
```
nicolaasmatthijs: Wouldn't this allow you to get users from a different app? We probably also need a test for that.

node_modules/gh-users/lib/api.js
```diff
@@ -90,13 +103,32 @@ var getUsers = module.exports.getUsers = function(ctx, app, query, limit, offset
         return callback({'code': 401, 'msg': 'You are not allowed to search/page users in this application'});
     }
 
+    /*!
+     * Hide personal information from a set of users
+     *
+     * @param  {Object} err   An error object, if any
+     * @param  {User[]} users The users to hide
```
nicolaasmatthijs: Needs extra spacing

nicolaasmatthijs: Also, can probably make this description a bit better

node_modules/gh-users/tests/test-search.js
```diff
@@ -226,4 +226,67 @@ describe('Users - Search', function() {
...
+    /**
+     * Assert whether personal information should be returned from the get/search users feed
+     *
+     * @param  {Object}         actor                   The object that holds the user and client to fetch the users
+     * @param  {Number}         app                     The id of the application to search in
+     * @param  {String}         [query]                 The query to filter users by
+     * @param  {Boolean}        expectPersonalInfo      Whether user objects should have personal information
+     * @param  {Function}       callback                Standard callback function
+     */
+    var expectUsers = function(actor, app, query, expectPersonalInfo, callback) {
```
nicolaasmatthijs: Shouldn't this be an assert function in the test util ?

node_modules/gh-users/tests/test-users.js
```diff
@@ -41,6 +41,82 @@ describe('Users', function() {
...
+    describe('Getting a user', function() {
+
+        /**
+         * Test that verifies that a user can be retrieved
+         */
+        it('verify a user can be retrieved', function(callback) {
+            TestsUtil.generateTestUsers(global.tests.apps.cam2014, 1, false, function(simon) {
+
+                UsersTestsUtil.assertGetUser(simon.client, simon.profile.id, function(user) {
+                    assert.strictEqual(user.displayName, simon.profile.displayName);
```
nicolaasmatthijs: Can we create an `assertUser` function?

node_modules/gh-users/tests/test-users.js
```diff
@@ -41,6 +41,82 @@ describe('Users', function() {
...
+
+        /**
+         * Test that verifies validation when retrieving a user
+         */
+        it('verify validation when retrieving a user', function(callback) {
+            TestsUtil.generateTestUsers(global.tests.apps.cam2014, 1, false, function(simon) {
+
+                UsersTestsUtil.assertGetUserFails(simon.client, 'Not a number', 400, function() {
+                    UsersTestsUtil.assertGetUserFails(simon.client, -1, 404, function() {
+                        UsersTestsUtil.assertGetUserFails(simon.client, 234234233, 404, function() {
```
nicolaasmatthijs: Should verify users from a different app here

node_modules/gh-users/tests/test-users.js
```diff
@@ -41,6 +41,82 @@ describe('Users', function() {
...
+
+        /**
+         * Test that verifies that personal information is not returned
+         */
+        it('verify personal information is not returned', function(callback) {
+            TestsUtil.generateTestUsers(global.tests.apps.cam2014, 2, false, function(simon, nico) {
+
+                // Simon can view his own information
+                UsersTestsUtil.assertGetUser(simon.client, simon.profile.id, function(user) {
+                    assert.strictEqual(user.displayName, simon.profile.displayName);
```
nicolaasmatthijs: All of these can be verified with an assertUser function

node_modules/gh-users/tests/util.js
```diff
@@ -38,6 +38,41 @@ var assertGetMe = module.exports.assertGetMe = function(client, callback) {
...
+ * @param  {RestClient}         client                          The REST client to make the request with
+ * @param  {Number}             id                              The id of the user to retrieve
+ * @param  {Function}           callback                        Standard callback function
+ * @param  {User}               callback.user                   The retrieved user
+ * @throws {AssertionError}                                     Error thrown when an assertion failed
+ */
+var assertGetUser = module.exports.assertGetUser = function(client, id, callback) {
+    client.user.getUser(id, function(err, user) {
+        assert.ok(!err);
+        assert.ok(user);
```
nicolaasmatthijs: Could optionally pass the expectedUser in here and compare all of the properties. Should verify that the basic properties are present regardless

nicolaasmatthijs:
Finished review. Back to @simong for follow-up.

node_modules/gh-core/lib/db.js
simong: It's probably ok, it's not like we're going to add lots of User preferences and if we do, the logic to hide them is right there. That said, I'll write a test to ensure only the properties we expect are returned.

node_modules/gh-series/tests/test-events.js
simong: Most of the organisers are plain-text authors here, there's nothing to index them on

node_modules/gh-users/tests/test-search.js
simong: I don't think so, its use is limited to this test file and it would make the `assertGetUsers` method signature a bit bloated I think

simong:

nicolaasmatthijs:
Exemplar review cycle I thought :) Merged

PR #46, opened 2015-01-15
Added some tests for user/admin authentication

simong:
Builds on https://github.com/fronteerio/grasshopper/pull/45

### DIFF
--- a/node_modules/gh-admins/lib/rest.js
+++ b/node_modules/gh-admins/lib/rest.js
@@ -16,6 +16,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+var _ = require('lodash');
+
 var GrassHopper = require('gh-core');
 
 var AdminsAPI = require('gh-admins');
@@ -53,7 +55,13 @@ GrassHopper.globalAdminRouter.on('get', '/api/admins', function(req, res) {
  * @Return      {Me}                                            The current global administrator
  */
 GrassHopper.globalAdminRouter.on('get', '/api/me', function(req, res) {
-    res.status(200).send(req.ctx.user);
+    var me = {
+        'anon': true
+    };
+    if (req.ctx.user) {
+        me = _.extend({'anon': false}, req.ctx.user.toJSON());
+    }
+    res.status(200).send(me);
 });
 
 /**
--- a/node_modules/gh-auth/tests/test-admins.js
+++ b/node_modules/gh-auth/tests/test-admins.js
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) 2014 "Fronteer LTD"
+ * Grasshopper Event Engine
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+var assert = require('assert');
+var TestsUtil = require('gh-tests');
+var UsersTestsUtil = require('gh-users/tests/util');
+
+var AuthTestsUtil = require('gh-auth/tests/util');
+
+describe('Local admin authentication', function() {
+
+    describe('Login', function() {
+
+        /**
+         * Test that verifies that an admin can log in
+         */
+        it('verify an admin can log in', function(callback) {
+            // Let a global admin create a user
+            TestsUtil.getAnonymousGlobalAdminRestClient(function(anonymousClient) {
+                // The anonymous client should be able to log in
+                AuthTestsUtil.assertLogin(anonymousClient, 'administrator', 'administrator', callback);
+            });
+        });
+
+        /**
+         * Test that verifies validation when logging in
+         */
+        it('verify validation when logging in', function(callback) {
+            // Get an anonymous client
+            TestsUtil.getAnonymousGlobalAdminRestClient(function(anonymousClient) {
+
+                // Using bad credentials should not allow a user to get authenticated
+                AuthTestsUtil.assertLoginFails(anonymousClient, 'administrator', null, function() {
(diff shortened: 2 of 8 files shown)


nicolaasmatthijs:
@simong : Would you be able to rebase this against master?

simong:
Rebased

node_modules/gh-admins/lib/rest.js
```diff
@@ -53,7 +55,13 @@ GrassHopper.globalAdminRouter.on('get', '/api/admins', function(req, res) {
  * @Return      {Me}                                            The current global administrator
  */
 GrassHopper.globalAdminRouter.on('get', '/api/me', function(req, res) {
-    res.status(200).send(req.ctx.user);
+    var me = {
```
nicolaasmatthijs: I would still prefer to have this in an API function. That would make it easier to return the me object when logging in or when updating the current user profile.

node_modules/gh-auth/tests/test-admins.js
```diff
@@ -0,0 +1,72 @@
...
+
+describe('Local admin authentication', function() {
+
+    describe('Login', function() {
+
+        /**
+         * Test that verifies that an admin can log in
+         */
+        it('verify an admin can log in', function(callback) {
+            // Let a global admin create a user
```
nicolaasmatthijs: Documentation doesn't seem correct

node_modules/gh-auth/tests/test-admins.js
```diff
@@ -0,0 +1,72 @@
...
+        /**
+         * Test that verifies validation when logging in
+         */
+        it('verify validation when logging in', function(callback) {
+            // Get an anonymous client
+            TestsUtil.getAnonymousGlobalAdminRestClient(function(anonymousClient) {
+
+                // Using bad credentials should not allow a user to get authenticated
+                AuthTestsUtil.assertLoginFails(anonymousClient, 'administrator', null, function() {
+                    AuthTestsUtil.assertLoginFails(anonymousClient, null, 'administrator', function() {
```
nicolaasmatthijs: Should we also verify that the admin credentials cannot be used to sign into an app?

node_modules/gh-auth/tests/test-users.js
```diff
@@ -0,0 +1,120 @@
...
+
+describe('Local user authentication', function() {
+
+    describe('Login', function() {
+
+        /**
+         * Test that verifies that a user can log in
+         */
+        it('verify a user can log in', function(callback) {
+            // Let a global admin create a user
```
nicolaasmatthijs: We have a util function for this

node_modules/gh-auth/tests/test-users.js
```diff
@@ -0,0 +1,120 @@
...
+         * Test that verifies validation when logging in
+         */
+        it('verify validation when logging in', function(callback) {
+            // Get an anonymous client for the app
+            TestsUtil.getAnonymousAppUserClient(global.tests.apps.oxford2014, function(anonymousClient) {
+
+                // Using bad credentials should not allow a user to get authenticated
+                AuthTestsUtil.assertLoginFails(anonymousClient, 'foo@foo.com', null, function() {
+                    AuthTestsUtil.assertLoginFails(anonymousClient, null, 'password', function() {
+                        AuthTestsUtil.assertLoginFails(anonymousClient, 'not an email', 'password', function() {
```
nicolaasmatthijs: Maybe we should also validate that valid credentials can't be used to sign into a different app

node_modules/gh-auth/tests/test-users.js
```diff
@@ -0,0 +1,120 @@
...
+                });
+            });
+        });
+
+        /**
+         * Test that verifies that a user needs the correct credentials to log in
+         */
+        it('verify a user needs the correct credentials to log in', function(callback) {
+            // Let a global admin create a user
+            var displayName = TestsUtil.generateString().toLowerCase();
```
nicolaasmatthijs: Can use util function for this

node_modules/gh-auth/tests/test-users.js
```diff
@@ -0,0 +1,120 @@
...
+            TestsUtil.generateTestUsers(global.tests.apps.cam2013, 1, false, function(simon) {
+                // Sanity-check the user is logged in
+                UsersTestsUtil.assertGetMe(simon.client, function(me) {
+                    assert.strictEqual(me.anon, false);
+
+                    // Log out
+                    AuthTestsUtil.assertLogout(simon.client, function() {
+
+                        // Verify the user is logged out
+                        UsersTestsUtil.assertGetMe(simon.client, function(me) {
```
nicolaasmatthijs: This part could be incorporated into the `AuthTestsUtil.assertLogout` function

node_modules/gh-auth/tests/util.js
```diff
@@ -24,6 +24,65 @@ var TestsUtil = require('gh-tests/lib/util');
...
+    });
+};
+
+/**
+ * Assert a user can not log in
+ *
+ * @param  {RestClient}     client                  The REST client to make the request with
+ * @param  {String}         username                The username for the user
+ * @param  {String}         password                The password for the user
+ * @param  {Function}       callback                Standard callback function
```
nicolaasmatthijs: Should we pass in the expected http response code here?

node_modules/gh-admins/lib/rest.js
nicolaasmatthijs: Thinking about this more, it would be good to add returning the me object after successful login to this PR

nicolaasmatthijs:

node_modules/gh-auth/tests/test-users.js
simong: Yes, but that generates a rest client and user object without providing the password. I think that in this case it's acceptable to create the user manually

simong: Never mind, the password is available at `user.client.options.password`

node_modules/gh-auth/tests/util.js
simong: When I was writing this, I was having trouble with passport return the proper response code, I think I left it out for that reason

simong: Oh, that's because passport always returns a 401, regardless of what we return

node_modules/gh-admins/lib/rest.js
simong: It's kinda fugly to write a generic `getMe` function and use it post-login. There's no context available yet at that time. There wouldn't be much benefit to it as the me feed only adds the `anon` key right now.

simong:

nicolaasmatthijs:
Looks good. Merged.

PR #47, opened 2015-01-15
Only retrieve that data we really need when getting a user’s calendar

simong:
Getting a calendar involves getting a ton of data from postgres and munging it in Sequelize. To avoid wasting cycles on stuff that we don't need, this query only selects those columns we actually need. 

This brought a massive improvement when doing performance tests. I will be revisiting the performance tests and try and build up a more realistic scenario and report some proper numbers back.

### DIFF
--- a/node_modules/gh-users/lib/internal/dao.js
+++ b/node_modules/gh-users/lib/internal/dao.js
@@ -370,20 +370,26 @@ var getUserCalendar = module.exports.getUserCalendar = function(user, includeCon
         dateTimeFilter.end = {'lte': end};
     }
 
+    // This query fetches such a massive amount of data, that the CPU usage
+    // goes through the roof when Sequelize serialises it. To minimise this,
+    // we retrieve only those columns we really need
+    var eventFields = ['id', 'displayName', 'start', 'end', 'location', 'organiserOther', 'updatedAt'];
+    var userFields = ['id', 'displayName'];
+
     var options = {
         // Adding `required` into the model inclusions forces Sequelize to use a
         // `left outer join` to connect the Serie/Events. If these were to be omitted,
         // Sequelize would use an `inner join` which would always result in 0 rows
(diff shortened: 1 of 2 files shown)


node_modules/gh-users/lib/internal/dao.js
```diff
@@ -370,20 +370,26 @@ var getUserCalendar = module.exports.getUserCalendar = function(user, includeCon
         dateTimeFilter.end = {'lte': end};
     }
 
+    // This query fetches such a massive amount of data, that the CPU usage
+    // goes through the roof when Sequelize serialises it. To minimise this,
+    // we retrieve only those columns we really need
```
nicolaasmatthijs: Can we add a verification to the tests that ensure that no other data is returned?

node_modules/gh-users/lib/internal/dao.js
```diff
@@ -370,20 +370,26 @@ var getUserCalendar = module.exports.getUserCalendar = function(user, includeCon
         dateTimeFilter.end = {'lte': end};
     }
 
+    // This query fetches such a massive amount of data, that the CPU usage
```
nicolaasmatthijs: Do you know how much of a difference this made during performance testing?

nicolaasmatthijs:

node_modules/gh-users/lib/internal/dao.js
simong: It's hard to compare tests as each run would add more series and modules to a user's calendar which would make subsequent runs slower. That said, between two runs I saw a drop of about 250ms on a mean of ~500ms. So roughly half.

I'll try to have a model loaded where each user has a set of subscribed series/modules and write a read-only performance test.

simong:

nicolaasmatthijs:
Looks good. Merged.

PR #50, opened 2015-01-22
Added logic to ensure that adding a serie to orgunit twice is ok

simong:
Also added a test for the events to serie use-case

### DIFF
--- a/node_modules/gh-orgunit/lib/api.js
+++ b/node_modules/gh-orgunit/lib/api.js
@@ -489,7 +489,7 @@ var addOrgUnitSeries = module.exports.addOrgUnitSeries = function(ctx, id, serie
     }
 
     // Ensure the organisational unit exists
-    OrgUnitDAO.getOrgUnit(id, false, function(err, orgUnit) {
+    OrgUnitDAO.getOrgUnit(id, true, function(err, orgUnit) {
         if (err) {
             return callback(err);
         }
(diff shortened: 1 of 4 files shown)


node_modules/gh-orgunit/lib/api.js
```diff
@@ -508,6 +508,16 @@ var addOrgUnitSeries = module.exports.addOrgUnitSeries = function(ctx, id, serie
                 return callback({'code': 400, 'msg': 'Some of the provided series belong to another application'});
             }
 
+            // Get all the series that are NOT part of the organisational unit yet
```
nicolaasmatthijs: yet --> already

node_modules/gh-orgunit/lib/api.js
```diff
@@ -508,6 +508,16 @@ var addOrgUnitSeries = module.exports.addOrgUnitSeries = function(ctx, id, serie
...
+            var newSeries = _.filter(series, function(serie) {
+                return !_.findWhere(orgUnit.Series, {'id': serie.id});
+            });
+
+            // If there are no new series, we can return early
+            if (_.isEmpty(newSeries)) {
```
nicolaasmatthijs: Do we have a db constraint on adding the same serie to an orgunit twice ?

nicolaasmatthijs:
Left a question and a super minor comment. Back to @simong

node_modules/gh-orgunit/lib/api.js
simong: Yes. Originally, I was going to do an if (err && err.name == UniqueIndexViolation) .. thing after the add, but it might get confusing when you're trying to add series to org units and only 1 is a duplicate

nicolaasmatthijs:
Looks good. Merged.

PR #56, opened 2015-02-02
Added an `academicYear` field to the config

simong:
We need to know what the term is for a given application so the UI can display the correct default calendar views/get the correct dates.

### DIFF
--- a/node_modules/gh-config/lib/api.js
+++ b/node_modules/gh-config/lib/api.js
@@ -230,10 +230,18 @@ var updateConfig = module.exports.updateConfig = function(ctx, appId, update, ca
     validator.check(_.keys(update).length, {'code': 400, 'msg': 'At least one config field needs to be specified'}).min(1);
     _.each(update, function(val, key) {
         validator.check(key, {'code': 400, 'msg': 'Unknown config field'}).isIn(ConfigDAO.EDITABLE_FIELDS);
+
         // Config fields that start with `allow` or `enable` should be boolean values
         if (key.indexOf('allow') === 0 || key.indexOf('enable') === 0) {
             val = GrasshopperUtil.getBooleanParam(val);
             validator.check(null, {'code': 400, 'msg': 'Boolean value expected for ' + key}).isBoolean(val);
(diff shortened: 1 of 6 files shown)


node_modules/gh-config/lib/api.js
```diff
@@ -230,12 +230,20 @@ var updateConfig = module.exports.updateConfig = function(ctx, appId, update, ca
...
+
         // Config fields that start with `allow` or `enable` should be boolean values
         if (key.indexOf('allow') === 0 || key.indexOf('enable') === 0) {
             val = GrasshopperUtil.getBooleanParam(val);
             validator.check(null, {'code': 400, 'msg': 'Boolean value expected for ' + key}).isBoolean(val);
             update[key] = val;
         }
+
+        // A valid integer must be provided for the `academic year` config field
+        if (key === 'academicYear') {
```
nicolaasmatthijs: Can anything be entered here, or should we enforce some sort of format (e.g. `2014`, `14`, etc.) ?

nicolaasmatthijs:

node_modules/gh-config/lib/api.js
```diff
@@ -230,12 +230,20 @@ var updateConfig = module.exports.updateConfig = function(ctx, appId, update, ca
...
+            val = GrasshopperUtil.getNumberParam(val);
+            validator.check(val, {'code': 400, 'msg': 'Integer value expected for ' + key}).isInt();
+            update[key] = val;
```
simong: It might be useful for other apps as well. As this can only be entered by administrators I don't think we need to put more reasoning behind it

node_modules/gh-config/lib/api.js
simong: Any number can be entered. This value will mostly be used to allow the UI to retrieve the correct set of static data (things such as term dates, etc..). It's up to them to keep it consistent (imho)

simong:
Back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #61, opened 2015-02-26
Need a way to know if a series has been borrowed from or to other modules.

simong:
The UI needs a way to determine whether a series has been borrowed from or to other modules.

e.g.,

- When deleting a series, the UI should be able to show that the series was borrowed from another module.
screen shot 2015-02-25 at 16 50 43

- When deleting a series, the UI should be able to list the modules that borrow this series.
screen shot 2015-02-25 at 16 50 11

### DIFF
--- a/node_modules/gh-rest/lib/rest/serie.js
+++ b/node_modules/gh-rest/lib/rest/serie.js
@@ -23,15 +23,19 @@ module.exports = function(client) {
      * Get a serie
      *
      * @param  {Number}         id                              The id of the serie to retrieve
+     * @param  {Boolean}        includeOrgUnits                 Whether the organisational units that this serie is a part of should be included in the response
      * @param  {Function}       callback                        Standard callback function
      * @param  {Object}         callback.err                    An error that occurred, if any
      * @param  {Object}         callback.body                   The JSON response from the REST API
      * @param  {Response}       callback.response               The response object as returned by requestjs
      * @see gh-series/lib/rest.js for more information
      */
-    client.serie.getSerie = function(id, callback) {
+    client.serie.getSerie = function(id, includeOrgUnits, callback) {
         var url = '/api/series/' + client.util.encodeURIComponent(id);
-        client._request(url, 'GET', null, null, callback);
+        var data = {
+            'includeOrgUnits': includeOrgUnits
+        };
+        client._request(url, 'GET', data, null, callback);
     };
 
     /**
--- a/node_modules/gh-rest/lib/rest/serie.js
+++ b/node_modules/gh-rest/lib/rest/serie.js
@@ -23,7 +23,7 @@ module.exports = function(client) {
      * Get a serie
      *
      * @param  {Number}         id                              The id of the serie to retrieve
-     * @param  {Boolean}        includeOrgUnits                 Whether the organisational units that this serie is a part of should be included in the response
+     * @param  {Boolean}        [includeOrgUnits]               Whether the organisational units that this serie is a part of should be included in the response. Defaults to `false`
      * @param  {Function}       callback                        Standard callback function
      * @param  {Object}         callback.err                    An error that occurred, if any
      * @param  {Object}         callback.body                   The JSON response from the REST API
--- a/node_modules/gh-series/lib/api.js
+++ b/node_modules/gh-series/lib/api.js
@@ -182,12 +182,12 @@ var updateSerie = module.exports.updateSerie = function(ctx, id, update, callbac
 /**
  * Get a serie
  *
- * @param  {Context}        ctx                 Standard context containing the current user and the current app
- * @param  {Number}         id                  The id of the serie to retrieve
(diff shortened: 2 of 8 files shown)


simong:
I've attached a commit that allows you to retrieve the organisational units where the serie is used in. If the number is more than 1 you could show the screen (it has all the org units, so you can show the list). Let me know if that works and I'll clean it up further.

bp323:
This works perfectly, thanks!

bp323:
One thing though, how can I determine which one was the original? It looks like the borrowed event is just duplicated and doesn't have any reference to the original one, unless I'm missing something.

simong:
I initially thought you only had to display a list where the series was being used. If we'll need the originating module, we'll need to add a join table

bp323:
We have 3 states:

- Delete the series --> can do as-is because if it's not borrowed, there won't be any orgunits coming back
- Delete the series that is borrowed by other modules --> can't do this as I don't know which of the orgunits coming back is the original
- Remove the series that you borrowed from another module --> can't do this as I don't know which of the orgunits coming back is the original

Let me know if it's not clear

simong:
@bp323 : I've rebased my branch, the order of the organisational units is now always in the same order as when the series was added to them. This means the first item will be the "original" module

So if you have `[ orgUnitA, orgUnitB ]`  it means the serie was first added to A and at a later time to B; meaning A was the "original" module.

FWIW, does the UI make a distinction between "Removing a series" and "Removing a series from a module (Unborrowing)" ?

bp323:
The UI makes the following distinctions:

- Not borrowed anywhere --> delete it from the system
- Borrowed in another module --> delete it form the system but give the user a heads up that it's being borrowed elsewhere
- Borrowed from another module --> remove it from the series but keep it intact in the system (unborrowing)

bp323:
FYI, the UI has everything it needs now and the PR is in at https://github.com/CUL-DigitalServices/grasshopper-ui/pull/205

simong:
Assigning to @nicolaasmatthijs for review

node_modules/gh-rest/lib/rest/serie.js
```diff
@@ -23,15 +23,19 @@ module.exports = function(client) {
      * Get a serie
      *
      * @param  {Number}         id                              The id of the serie to retrieve
+     * @param  {Boolean}        includeOrgUnits                 Whether the organisational units that this serie is a part of should be included in the response
```
nicolaasmatthijs: This is optional

node_modules/gh-series/lib/api.js
```diff
@@ -184,21 +184,27 @@ var updateSerie = module.exports.updateSerie = function(ctx, id, update, callbac
  *
  * @param  {Context}        ctx                 Standard context containing the current user and the current app
  * @param  {Number}         id                  The id of the serie to retrieve
+ * @param  {Boolean}        includeOrgUnits     Whether the organisational units that this serie is a part of should be included in the response
```
nicolaasmatthijs: Optional

node_modules/gh-series/lib/internal/dao.js
```diff
@@ -78,15 +78,21 @@ var updateSerie = module.exports.updateSerie = function(serie, update, callback)
  * Get a serie
  *
  * @param  {Number}         id                          The id of the serie to retrieve
+ * @param  {Boolean}        includeOrgUnits             Whether the organisational units that this serie is a part of should be included in the response
```
nicolaasmatthijs: Same here

node_modules/gh-series/lib/rest.js
```diff
@@ -30,12 +30,13 @@ var SeriesAPI = require('./api');
  * @Method      GET
  * @Path        /series/{id}
  * @PathParam   {number}            id                  The id of the event series to retrieve
+ * @QueryParam  {boolean}           includeOrgUnits     Whether the organisational units that this serie is a part of should be included in the response
```
nicolaasmatthijs: And here

node_modules/gh-series/tests/test-crud.js
```diff
@@ -156,15 +157,21 @@ describe('Series', function() {
...
                 // Invalid serie id
-                SeriesTestUtil.assertGetSerieFails(simon.client, 'not a number', 400, function() {
+                SeriesTestUtil.assertGetSerieFails(simon.client, 'not a number', false, 400, function() {
 
                     // Unknown serie id
-                    SeriesTestUtil.assertGetSerieFails(simon.client, -1, 404, function() {
-                        return callback();
+                    SeriesTestUtil.assertGetSerieFails(simon.client, -1, false, 404, function() {
+
+                        // Invalid boolean flag
```
nicolaasmatthijs: `Invalid include orgunit boolean flag`

nicolaasmatthijs:
Looks good. Left some minor feedback items. Back to @simong for follow-up.

simong:

nicolaasmatthijs:
Looks good. Merged.

PR #80, opened 2015-03-19
Borrowing

simong:
Simpler version of borrowing.

The first organisational unit is now  the "original" one.

### DIFF
--- a/node_modules/gh-orgunit/tests/test-orgunit.js
+++ b/node_modules/gh-orgunit/tests/test-orgunit.js
@@ -908,6 +908,49 @@ describe('Organisational units', function() {
                 });
             });
         });
+
+        /**
+         * Test that verifies that the organisational unit a series is borrowed from is included in the response
+         */
+        it('verify the organisational unit a series is borrowed from is included in the response', function(callback) {
+            // Generate a structure where a series is borrowed under multiple organisational units
+            OrgUnitTestUtil.createSimpleTree(global.tests.admins.cam2013.client, function(courseUnit, subjectUnit, partUnit, moduleUnit1, series) {
+                OrgUnitTestUtil.assertCreateOrgUnit(global.tests.admins.cam2013.client, '#2', 'module', null, function(moduleUnit2) {
+                    OrgUnitTestUtil.assertCreateOrgUnit(global.tests.admins.cam2013.client, '#2', 'module', null, function(moduleUnit3) {
+                        var seriesToAdd = series[0];
+                        OrgUnitTestUtil.assertAddOrgUnitSeries(global.tests.admins.cam2013.client, moduleUnit2.id, [seriesToAdd.id], function() {
+                            OrgUnitTestUtil.assertAddOrgUnitSeries(global.tests.admins.cam2013.client, moduleUnit3.id, [seriesToAdd.id], function() {
+
+                                OrgUnitTestUtil.assertGetOrgUnits(global.tests.admins.cam2013.client, null, true, null, null, 'module', function(orgUnits) {
+
+                                    // There's no borrowing taking place in module 1 as it was the first organisational
+                                    // unit to add the series. The `borrowedFrom` key should not be present
+                                    var moduleUnit1Response = _.find(orgUnits.results, {'id': moduleUnit1.id});
+                                    var series = _.find(moduleUnit1Response.Series, {'id': seriesToAdd.id});
+                                    assert.ok(!series.borrowedFrom);
+
+                                    // Assert that module 2 is borrowing the series from module 1
+                                    var moduleUnit2Response = _.find(orgUnits.results, {'id': moduleUnit2.id});
+                                    series = _.find(moduleUnit2Response.Series, {'id': seriesToAdd.id});
+                                    assert.ok(series.borrowedFrom);
+                                    assert.strictEqual(series.borrowedFrom.id, moduleUnit1.id);
+                                    var allowedKeys = ['id', 'displayName', 'ParentId'].sort();
+                                    assert.deepEqual(_.keys(series.borrowedFrom).sort(), allowedKeys);
+
+                                    // Assert that module 2 is borrowing the series from module 1
+                                    var moduleUnit3Response = _.find(orgUnits.results, {'id': moduleUnit3.id});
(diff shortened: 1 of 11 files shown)


simong:
Got the OK from the Cambridge guys. This is ready for review.

node_modules/gh-orgunit/tests/test-orgunit.js
```diff
@@ -908,6 +908,49 @@ describe('Organisational units', function() {
...
+                        var seriesToAdd = series[0];
+                        OrgUnitTestUtil.assertAddOrgUnitSeries(global.tests.admins.cam2013.client, moduleUnit2.id, [seriesToAdd.id], function() {
+                            OrgUnitTestUtil.assertAddOrgUnitSeries(global.tests.admins.cam2013.client, moduleUnit3.id, [seriesToAdd.id], function() {
+
+                                OrgUnitTestUtil.assertGetOrgUnits(global.tests.admins.cam2013.client, null, true, null, null, 'module', function(orgUnits) {
+
+                                    // There's no borrowing taking place in module 1 as it was the first organisational
+                                    // unit to add the series. The `borrowedFrom` key should not be present
+                                    var moduleUnit1Response = _.find(orgUnits.results, {'id': moduleUnit1.id});
+                                    var series = _.find(moduleUnit1Response.Series, {'id': seriesToAdd.id});
```
nicolaasmatthijs: Would it be worth extending `assertSeries` with a `expectedBorrowedFrom` parameter?

node_modules/gh-orgunit/tests/test-orgunit.js
```diff
@@ -908,6 +908,49 @@ describe('Organisational units', function() {
...
+
+                                    // Assert that module 2 is borrowing the series from module 1
+                                    var moduleUnit2Response = _.find(orgUnits.results, {'id': moduleUnit2.id});
+                                    series = _.find(moduleUnit2Response.Series, {'id': seriesToAdd.id});
+                                    assert.ok(series.borrowedFrom);
+                                    assert.strictEqual(series.borrowedFrom.id, moduleUnit1.id);
+                                    var allowedKeys = ['id', 'displayName', 'ParentId'].sort();
+                                    assert.deepEqual(_.keys(series.borrowedFrom).sort(), allowedKeys);
+
+                                    // Assert that module 2 is borrowing the series from module 1
```
nicolaasmatthijs: Isn't this essentially the same test as the one above ?

node_modules/gh-series/lib/api.js
```diff
@@ -201,18 +201,27 @@ var getSerie = module.exports.getSerie = function(ctx, id, includeOrgUnits, call
...
         }
 
-        return callback(null, serie);
+        // Check whether the current user can manage the series
+        SeriesAuthz.canManageSeries(ctx, series, function(err, canManage) {
+            if (err) {
+                return callback(err);
+            }
+
+            series.canManage = canManage;
```
nicolaasmatthijs: Do we need to extends the tests/test utilities to validate this property ?

node_modules/gh-orgunit/tests/test-orgunit.js
simong: Yes, it's just double-checking that it can deal with more than 1 borrower

node_modules/gh-series/tests/test-crud.js
```diff
@@ -178,22 +178,41 @@ describe('Series', function() {
...
                         // Assert that it can't be retrieved on the 2014 app
-                        SeriesTestUtil.assertGetSerieFails(nico.client, createdSerie.id, false, 401, function() {
+                        SeriesTestUtil.assertGetSerieFails(nico.client, createdSeries.id, false, 401, function() {
 
                             // Assert a global admin can retrieve it
                             TestsUtil.getGlobalAdminRestClient(function(adminClient) {
-                                SeriesTestUtil.assertGetSerie(adminClient, createdSerie.id, false, createdSerie, function(retrievedSerie) {
-                                    return callback();
+                                SeriesTestUtil.assertGetSerie(adminClient, createdSeries.id, false, createdSeries, function(series) {
+                                    assert.strictEqual(series.canManage, true);
```
nicolaasmatthijs: Might be worth making this check part of the test util, so we can at least test that it's always returned and a boolean

node_modules/gh-series/tests/test-crud.js
```diff
@@ -246,6 +265,43 @@ describe('Series', function() {
                 });
             });
         });
+
+        /**
+         * Test that verifies that the organisational unit a series is borrowed from is indicated when getting a series and including the organisational units
```
nicolaasmatthijs: As this is already under the `Getting a series` block, this can be changed to `Test that verifies that the organisational unit a series is borrowed from is returned when including the organisational units`

nicolaasmatthijs:
From #81:

```
I have 5 organisational units (module type), each with the same parent. 2 of those organisational units 
have event series associated with them.

When I request /api/orgunit?app=1&includeSeries=false&parent=1112&type=module I get the correct results: 5 organisational units without series included.

When I request /api/orgunit?app=1&includeSeries=true&parent=1112&type=module I get incorrect results (note the only difference being includeSeries). I'm expecting 5 organisational units to return, 2 of them having an Array of event series, 3 of them no event series. However, what I get is only 2 organisational units, the 2 that have event series associated to them. I would expect the 3 without event series to show up as well.
```

We should add a test for this as well.

nicolaasmatthijs:

node_modules/gh-orgunit/tests/test-orgunit.js
simong: I don't think so, we only return it:
 - from `/api/series/:id` if the `includeOrgUnits` flag is set to `true`
 - from /api/orgUnits if the `includeSeries` flag is set to `true`

node_modules/gh-series/lib/api.js
simong: There is test coverage for this under the `verify authorization when getting a series` test.
I'll extend the `assertGetSerie` function so it does the canManage check.

simong:
Pushed some follow-up, back to @nicolaasmatthijs

nicolaasmatthijs:
Only comment-related feedback left. I'll follow up and merge.

nicolaasmatthijs:
Followed up and merged

PR #82, opened 2015-03-23
Rename Google Analytics configuration variables

bp323:
The UI will be using segment.io to pipe data through to various analytics platforms. I've renamed the GA configuration variables to more generic names.


simong:
Thanks! I've pushed a small follow-up and merged this in.

PR #84, opened 2015-04-01
Cambridge patterns aggregation can be removed

simong:
Everything under gh-series/lib/internal/patterns can be ripped out as it's no longer needed


simong:
Not *everything* could be ripped out as we still have to aggregate the locations and organisers, but at least the Cambridge specific logic can be removed

simong:
I'm going to close this as it's actually kind of useful in the import scripts. I'll move them into those scripts in that PR.

PR #86, opened 2015-04-03
Implemented the import users endpoint

simong:
This allows a global administrator to upload a CSV file and import a set of users. The request parameters AND csv file will be validated prior to actual
user creation/updating. The actual creation/updating happens async from the request.

This also brings in busboy as the new HTTP body parser


node_modules/gh-core/lib/server.js
```diff
@@ -105,14 +105,15 @@ var setUpServer = module.exports.setUpServer = function(port, config) {
...
      *
      * A maximum limit of 250kb is imposed for `urlencoded` and `application/json` requests. This limit only
      * applies to the *incoming request data*. If the client needs to send more than 250kb, it should consider
      * using a proper multipart form request
      */
-    app.use(bodyParser.urlencoded({'limit': '250kb', 'extended': true}));
-    app.use(bodyParser.json({'limit': '250kb'}));
+    busboy.extend(app, {
+        'upload': true
+    });
```
mrvisser: Do the 250kb limits apply now that we've switched to `busboy`?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+    validator.check(null, {'code': 400, 'msg': 'A valid profile update flag must be provided'}).isBoolean(forceProfileUpdate);
+    if (validator.hasErrors()) {
+        _cleanUpCSVFile(file, function() {
+            return callback(validator.getFirstError());
+        });
+        return;
+    }
+
+    // Ensure the application exists
+    AppsAPI.getApp(ctx, appId, function(err, app) {
```
mrvisser: `AppsAPI.getApp` sounds like something we would want to roll into the `ctx`

mrvisser: e.g., `ctx.app()`

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+        if (err) {
+            return callback(err);
+        }
+
+        // Ensure the current user can import users
+        UsersAuthz.canImportUsers(ctx, app, function(err, canImportUsers) {
+            if (err) {
+                return callback(err);
+            } else if (!canImportUsers) {
+                log().warn({'app': appId, 'actor': ctx.user.id}, 'Unauthorized attempt at importing users');
```
mrvisser: Any particular to log a warning as opposed to just logging the request error as usual?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                // Regardless of the error, get rid of the CSV file first
+                _cleanUpCSVFile(file, function() {
+                    if (err) {
+                        log().error({'err': err, 'app': appId}, 'Failed to parse CSV file when importing users');
+                        return callback({'code': 500, 'msg': err.message});
+                    }
+
+                    // Validate each record before attempting to create users. Although this makes
+                    // the request a bit slower, it prevents us from ending up with only half of the
+                    // CSV file's users
```
mrvisser: I don't think validating up front makes it slower

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                    _.each(users, function(user) {
+                        validator.check(user[0], {'code': 400, 'msg': 'Missing display name for: ' + user[1]}).notEmpty();
+                        if (authenticationStrategy === 'local') {
+                            validator.check(user[1], {'code': 400, 'msg': 'A valid email address must be provided for: ' + user[0]}).isEmail();
+                            validator.check(user[2], {'code': 400, 'msg': 'A valid password must be provided for: ' + user[0]}).notEmpty();
+                            validator.check(user[2], {'code': 400, 'msg': 'A valid password must be provided for: ' + user[0]}).len(6);
+                        } else {
+                            validator.check(user[2], {'code': 400, 'msg': 'A valid shibboleth id must be provided for: ' + user[0]}).notEmpty();
```
mrvisser: Can we use `util.format` for all these?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                        // Indicate that all the users from the CSV file have been processed
+                        UsersAPI.emit('postImportUsers', err);
+                    });
+                });
+            });
+
+            // The CSV module works with streams, so get a readable stream to the uploaded CSV file
+            // and pipe it to the CSV parser
+            var input = fs.createReadStream(file.file);
+            input.pipe(parser);
```
mrvisser: I think `fs.createReadStream(file.file).pipe(parser);` is nicer for this, if you agree

node_modules/gh-users/lib/api.js
simong: FWIW, we have the *current* application available as a property of the context at `ctx.app`

mrvisser: I see, so it's possible to send a request to app 1 that imports users into app 2?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                    // Indicate the request is valid and we will start actually importing users
+                    UsersAPI.emit('preImportUsers');
+
+                    // If the CSV parse was successful, we call the callback to prevent the request from timing out
+                    // whilst the users are being loaded
+                    callback();
+
+                    // Import the users
+                    log().info({
+                        'app': appId,
```
mrvisser: `'appId'` as the property key

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                        'authenticationStrategy': authenticationStrategy
+                    }, 'Starting user import from CSV');
+                    _importUsers(appId, authenticationStrategy, forceProfileUpdate, users, function(err) {
+                        if (!err) {
+                            log().info({
+                                'app': appId,
```
mrvisser: `'appId'`

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+ */
+var _importUsers = function(app, authenticationStrategy, forceProfileUpdate, users, callback) {
+    if (_.isEmpty(users)) {
+        return callback();
+
+    // Add a progress log statement every 50 imported users
+    } else if (users.length % 50 === 0) {
+        log().info({
+            'authenticationStrategy': authenticationStrategy,
+            'app': app
```
mrvisser: `'appId'`

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+
+        // Recursively import the next user (if any)
+        _importUsers(app, authenticationStrategy, forceProfileUpdate, users, callback);
+    });
+};
+
+/**
+ * Import a user
+ *
+ * @param  {Number}     app                         The id of the application to import the user into
```
mrvisser: The variable name here should be `appId` I think

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+ * @param  {String}     authenticationStrategy      The authentication strategy for the user. One of `local` or `shibboleth`
+ * @param  {Boolean}    [forceProfileUpdate]        Whether the user information should be updated, even when other user information is already present
+ * @param  {Object}     user                        The user to import
```
mrvisser: It seems like this might be an array of strings, not an object? If feasible, can you enumerate what we expect at what index?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+        UsersAuthz.canImportUsers(ctx, app, function(err, canImportUsers) {
+            if (err) {
+                return callback(err);
+            } else if (!canImportUsers) {
+                log().warn({'app': appId, 'actor': ctx.user.id}, 'Unauthorized attempt at importing users');
+                return callback({'code': 401, 'msg': 'You are not allowed to import users'});
+            }
+
+            // Construct the CSV parser
+            var parser = csv.parse({'trim': true}, function(err, users) {
```
mrvisser: Can we be more clear that these are `userRows`, and only switch to using pure `user` terminology if we're dealing with a `user` model object.

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                // Regardless of the error, get rid of the CSV file first
+                _cleanUpCSVFile(file, function() {
+                    if (err) {
+                        log().error({'err': err, 'app': appId}, 'Failed to parse CSV file when importing users');
```
mrvisser: `'appId'`

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                        return callback({'code': 500, 'msg': err.message});
+                    }
+
+                    // Validate each record before attempting to create users. Although this makes
+                    // the request a bit slower, it prevents us from ending up with only half of the
+                    // CSV file's users
+                    _.each(users, function(user) {
```
mrvisser: `userRows` / `userRow`

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+    }
+};
+
+/**
+ * Import a set of users
+ *
+ * @param  {Number}     app                         The id of the application to import the users into
+ * @param  {String}     authenticationStrategy      The authentication strategy for the user. One of `local` or `shibboleth`
+ * @param  {Boolean}    [forceProfileUpdate]        Whether the user information should be updated, even when other user information is already present
+ * @param  {Object[]}   users                       The users to import
```
mrvisser: I think this is a `String[][]`, and should be called `userRows`

node_modules/gh-users/lib/api.js
mrvisser: Also, `userRow`

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+    }
+
+    var credentials = {
+        'strategy': authenticationStrategy
+    };
+    var identifier = null;
+    if (authenticationStrategy === 'local') {
+        identifier = user[1];
+        credentials.password = user[2];
+    } else if (authenticationStrategy === 'shibboleth') {
```
mrvisser: Should there be "constants" available for `'local'` and `'shibboleth'` ?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,243 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+    // trigger a constraint violation if we just passed it on. To avoid that, we only set the email
+    // address if one was actually provided
+    if (user[1]) {
+        userProfile.email = user[1];
+    }
+
+    var credentials = {
+        'strategy': authenticationStrategy
+    };
+    var identifier = null;
```
mrvisser: It's kind of confusing that users have an `id`, which is their identifier, and also `identifier` which is really their identifier in an external system (e.g., `externalId`).

node_modules/gh-users/lib/rest.js
```diff
@@ -340,18 +340,34 @@ GrassHopper.appRouter.on('post', '/api/users', function(req, res) {
...
+ * of your CSV file this might take a couple of minutes to complete
+ *
  * @Server      admin
  * @Method      POST
  * @Path        /users/import
  * @FormParam   {string}            authenticationStrategy  The authentication strategy for the user    [local,shibboleth]
  * @FormParam   {File}              file                    The CSV file to import
- * @FormParam   {number}            tenant                  The id of the tenant to which the users should be imported
- * @FormParam   {number[]}          [app]                   The id(s) of the app(s) to which the users should be associated
+ * @FormParam   {number}            [app]                   The id of the application into which the users should be imported. Defaults to the current application
```
mrvisser: Do we use `app` as the parameter for the app id in lots of other places? If not, I'd suggest making this `appId` if possible.

node_modules/gh-users/lib/rest.js
```diff
@@ -340,18 +340,34 @@ GrassHopper.appRouter.on('post', '/api/users', function(req, res) {
...
  * @FormParam   {boolean}           [forceProfileUpdate]    Whether the user information should be updated, even when other user information is already present. Defaults to `false`
  * @Return      {void}
  */
 GrassHopper.globalAdminRouter.on('post', '/api/users/import', function(req, res) {
-    res.sendStatus(501);
+    UsersAPI.importUsers(req.ctx, req.files.file, req.body.app, req.body.authenticationStrategy, req.body.forceProfileUpdate, function(err) {
+        if (err) {
+            return res.status(err.code).send(err.msg);
```
mrvisser: `res.send(err.code, err.msg);`

node_modules/gh-users/lib/rest.js
```diff
@@ -340,18 +340,34 @@ GrassHopper.appRouter.on('post', '/api/users', function(req, res) {
...
+        }
+
+        return res.status(200).send({});
```
mrvisser: `res.send(200, {});`

node_modules/gh-users/tests/test-import.js
```diff
@@ -0,0 +1,302 @@
+/**
+ * Copyright (c) 2014 "Fronteer LTD"
```
mrvisser: `2015`

node_modules/gh-users/tests/test-import.js
```diff
@@ -0,0 +1,302 @@
...
+var fs = require('fs');
+
+var AuthTestsUtil = require('gh-auth/tests/util');
+var ConfigTestsUtil = require('gh-config/tests/util');
+var RestAPI = require('gh-rest');
+var TestsUtil = require('gh-tests/lib/util');
+
+var UsersTestsUtil = require('./util');
+
+describe('Users - import', function() {
```
mrvisser: We could just do 2 describes, one for `Users` and one for `Import`, rather than creating a text-based naming convention.

node_modules/gh-users/lib/rest.js
simong: Unfortunately, that's been deprecated

node_modules/gh-users/lib/rest.js
simong: Yes, we tend to not suffix parameter names with `Id` in our REST endpoints. For non-REST-parameters I try to stick to `appId` when dealing with the application id and `app` when dealing with the application object.

node_modules/gh-users/lib/api.js
simong: It's been a practice we've been doing in all our API functions whenever a 401 pops up. I suppose we could consolidate this in some middleware.

node_modules/gh-users/tests/test-import.js
```diff
@@ -0,0 +1,302 @@
...
+                    stream = fs.createReadStream(__dirname + '/data/users-shib.csv');
+                    UsersTestsUtil.assertImportUsers(globalAdminClient, stream, app.id, 'shibboleth', false, function() {
+
+                        // Assert the users were imported
+                        UsersTestsUtil.assertGetUsers(globalAdminClient, app.id, 'Christina McChristina', null, null, function(users) {
+                            var user = _.find(users.results, {'displayName': 'Christina McChristina'});
+                            assert.ok(user);
+                            assert.strictEqual(user.displayName, 'Christina McChristina');
+                            assert.strictEqual(user.email, 'christina@example.com');
+                            UsersTestsUtil.assertUser(user, null, true);
```
mrvisser: It might be too late to change this, but a function like this is really hard to read without having all parameters memorized (which is never, for me), because it doesn't indicate anything about what it is asserting. It would be nice if we just had methods that asserted something specific, like:

`assertUserEquals`
`assertUserVisible`

mrvisser: Another option:

```
assertUser(user, {
  'equalsUser': anotherUser,
  'canViewAllFields': true
});
```

At least that way the properties give an indication of what's being asserted.

node_modules/gh-users/tests/test-import.js
```diff
@@ -0,0 +1,302 @@
...
+                                'displayname': 'A different value from the CSV'
+                            };
+                            AuthTestsUtil.assertShibbolethLogin(app, 'shibNewland', attributes, null, function(restClient) {
+
+                                // Verify we're authenticated
+                                UsersTestsUtil.assertGetMe(restClient, function(me) {
+                                    assert.strictEqual(me.id, user.id);
+                                    assert.strictEqual(me.displayName, user.displayName);
+                                    assert.strictEqual(me.email, user.email);
+                                    assert.strictEqual(me.shibbolethId, user.shibbolethId);
```
mrvisser: Should `assertUser` / `assertUserEquals` be used for this?

node_modules/gh-users/lib/api.js
mrvisser: Yea, or just HTTP logs.

mrvisser:
Looks really good overall, back to @simong for some followup.

node_modules/gh-users/lib/api.js
simong: Not really as the endpoint is only exposed on the global admin router.

node_modules/gh-users/lib/api.js
simong: I'm not entirely sure what you want me to do about this, but I'll try to rename the local variable here.

mrvisser: I think it's good if our variable nomenclature (down to the DAO) could indicate `externalId`. Ideally we could split it out of the user table as well rather than overloading it, but that can hold off until we introduce another authentication strategy that introduces another `externalId`.

node_modules/gh-users/tests/test-import.js
simong: /cc @nicolaasmatthijs any preference?

nicolaasmatthijs: I agree with the readability issue. I've just started switching to the second approach in a different project:

```
assertX(x, {'expectedUser': user, ...});
```

I think that's the best combination of readability and flexibility

simong:
Followed-up on review, back to @mrvisser

node_modules/gh-core/lib/server.js
```diff
@@ -61,6 +61,44 @@ var setUpServer = module.exports.setUpServer = function(port, config) {
...
+    // Store the app on which the request was made on the request object
+    app.use(function(req, res, next) {
+        // Pass in a global admin server object when the request was made
+        // on the global admin host
+        if (config.servers.adminHostname === req.hostname) {
+            var globalAdmin = {
+                'displayName': 'Global admin server',
+                'host': req.hostname,
+                'isGlobalAdmin': true
+            };
```
mrvisser: Do we have constructors to create these objects instead? Also, what if the AppsAPI took care of doing these admin and shib hostname checks? Then the logic is centralized in a re-usable place.

Something like a single `AppsAPI.getAppByHost(...)` which will do the shib and admin host check, then fall back to the persisted app for the host, if any.

mrvisser: Or maybe a Util instead of an API if we can't get access to a context. But the DAO would preferably not hide this kind of logic.

node_modules/gh-core/tests/test-core.js
```diff
@@ -0,0 +1,111 @@
+/**
+ * Copyright (c) 2014 "Fronteer LTD"
```
mrvisser: 2015

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,242 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+                log().warn({'appId': appId, 'actor': ctx.user.id}, 'Unauthorized attempt at importing users');
+                return callback({'code': 401, 'msg': 'You are not allowed to import users'});
+            }
+
+            // Construct the CSV parser
+            var parser = csv.parse({'trim': true}, function(err, userRows) {
+                // Regardless of the error, get rid of the CSV file first
+                _cleanUpCSVFile(file, function() {
+                    if (err) {
+                        log().error({'err': err, 'app': appId}, 'Failed to parse CSV file when importing users');
```
mrvisser: `'appId': appId`

node_modules/gh-users/lib/api.js
mrvisser: Were you intending to rename references to this "identifier" as "externalId"?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,242 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+    });
+};
+
+/**
+ * Import a user
+ *
+ * @param  {Number}     appId                       The id of the application to import the user into
+ * @param  {String}     authenticationStrategy      The authentication strategy for the user. One of `local` or `shibboleth`
+ * @param  {Boolean}    [forceProfileUpdate]        Whether the user information should be updated, even when other user information is already present
+ * @param  {String[]}   userRow                     The user row to import
```
mrvisser: Is it feasible to indicate what value should be at what index of the row in the documentation?

node_modules/gh-users/lib/api.js
```diff
@@ -324,6 +335,242 @@ var updateAdminStatus = module.exports.updateAdminStatus = function(ctx, userId,
...
+    }
+
+    var credentials = {
+        'strategy': authenticationStrategy
+    };
+    var identifier = null;
+    if (authenticationStrategy === 'local') {
+        identifier = userRow[1];
+        credentials.password = userRow[2];
+    } else if (authenticationStrategy === 'shibboleth') {
```
mrvisser: Should there be constants available for `'local'` and `'shibboleth'`

mrvisser:
I've added some more comments, back to @simong for followup.

simong:
Pushed follow-up, back to @mrvisser

mrvisser:
:+1: !

PR #90, opened 2015-04-09
(DeleteGlobalAdmins) Add API endpoints to delete a global admin

mrvisser:
Todo: Add a few validation tests

### DIFF
--- a/node_modules/gh-admins/tests/test-admins.js
+++ b/node_modules/gh-admins/tests/test-admins.js
@@ -86,5 +86,25 @@ describe('Admins', function() {
                 });
             });
         });
+
+        /**
+         * Test that verifies validation of the delete global admin parameters
+         */
+        it('verify validation of deleting a global admin', function(callback) {
+            // Create a global administrator to sanity check deletes with
+            TestsUtil.getGlobalAdminRestClient(function(globalAdminClient) {
+                var username = TestsUtil.generateTestUserId();
+                AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, username, username, function(createdGlobalAdmin) {
+
+                    AdminsTestsUtil.assertDeleteGlobalAdminFails(globalAdminClient, 'not an id', 400, function() {
+                        AdminsTestsUtil.assertDeleteGlobalAdminFails(globalAdminClient, '-1', 404, function() {
+
+                            // Sanity check we can delete a global admin
+                            return AdminsTestsUtil.assertDeleteGlobalAdmin(globalAdminClient, createdGlobalAdmin.id, callback);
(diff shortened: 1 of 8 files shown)


mrvisser:
This is ready for review. Assigning to @simong

mrvisser:
FYI, I'll be following up with Authz and unit tests for admins

node_modules/gh-admins/tests/test-admins.js
```diff
@@ -0,0 +1,110 @@
...
+        /**
+         * Test that verifies a global admin can delete their own user account
+         */
+        it('verify a global admin can delete their own user account', function(callback) {
+            TestsUtil.getGlobalAdminRestClient(function(globalAdminClient) {
+                var username = TestsUtil.generateTestUserId();
+                AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, username, username, function(createdGlobalAdmin) {
+
+                    // Delete the global admin, ensuring they no longer exist
+                    return AdminsTestsUtil.assertDeleteGlobalAdmin(globalAdminClient, createdGlobalAdmin.id, callback);
```
simong: If you delete your own account, should you be logged out?

mrvisser: It would probably make sense to make it so that a session that is for a user id that doesn't exist is equivalent to an anonymous session.

node_modules/gh-rest/lib/rest/admin.js
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Get a list of global administrators
+     *
+     * @param  {Object}             [opts]                  Optional properties
+     * @param  {Number}             [opts.limit]            The maximum number of results to retrieve. Default: 10
+     * @param  {Number}             [opts.offset]           The paging number of the results to retrieve
+     * @param  {Function}           callback                Standard callback function
+     * @param  {Object}             callback.err            An error that occurred, if any
+     * @param  {GlobalAdminList}    callback.globalAdmins   The requested list of global administrators
```
simong: FWIW, for the gh-rest we usually document the callback parameters like so:
```
     * @param  {Function}       callback                Standard callback function
     * @param  {Object}         callback.err            An error that occurred, if any
     * @param  {Object}         callback.body           The JSON response from the REST API
     * @param  {Response}       callback.response       The response object as returned by requestjs
     * @see gh-admins/lib/rest.js for more information
```

No need to change these though

node_modules/gh-admins/tests/test-admins.js
simong: I think I made this change in one of my upcoming PRs

PR #95, opened 2015-04-14
Leave poor grasshopper.com alone

mrvisser:
I have a thing using non-obvious sample URLs as sample/default URLs, especially when they resolve to something on the internet, and especially especially when they resolve to what looks to be valid existing company domains.

We can do all the things we need with `/etc/hosts` with `*.local` domains which are more common for things that we intend to resolve locally (e.g., default URLs), so it's better a bit more clear about the intention when we default to `*.local`s.

### DIFF
--- a/config.js
+++ b/config.js
@@ -86,10 +86,10 @@ config.cookie = {
  * @property  {String}      shibbolethSPHost        The hostname on which the Shibboleth Service Provider software will be made available
  */
 config.servers = {
-    'adminHostname': 'admin.grasshopper.com',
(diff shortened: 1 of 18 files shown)


mrvisser:
I've also made it so that the hostnames specified in `config.js` don't break any tests, and it's now not necessary for the tests to override the configured hosts for anything.

mrvisser:
Assigned to @simong for review

simong:
PR looks good. Feel free to merge after rebasing and fixing up/explaining that one comment

simong:
Merged this into master manually

PR #98, opened 2015-04-20
Made the hostname on which the webservers are spun up configurable

simong:


config.js
```diff
@@ -80,12 +80,14 @@ config.cookie = {
...
  * The servers related configuration
  *
+ * @property  {String}      serverHostname          The hostname on which the web servers should be listening. See https://nodejs.org/api/http.html#http_server_listen_port_hostname_backlog_callback
  * @property  {String}      adminHostname           The hostname on which the admin UI will be made available
  * @property  {Number}      adminPort               The port on which the admin API endpoints will be made available
  * @property  {Number}      appsPort                The port on which the regular application API endpoints will be made available
  * @property  {String}      shibbolethSPHost        The hostname on which the Shibboleth Service Provider software will be made available
  */
 config.servers = {
+    'webServerHostname': '0.0.0.0',
```
mrvisser: 'Hostname' is confusing here because we use that to say the **external** host at which to access the admin server.

It might be worth to first do some investigation into what other server platforms (redis, elasticsearch, nginx, apache, rabbitmq, etc...) use as terminology to specify on which interface the platform will "listen" for incoming requests. That way it will be more self-explanatory to system administrators.

mrvisser:
It would be worth making other hardcoded areas use the dynamic values as well, both listen interface and port:

```
$ git grep "'localhost"
node_modules/gh-tests/lib/util.js:        'host': 'localhost:2000',
node_modules/gh-tests/lib/util.js:        'host': 'localhost:2000',
node_modules/gh-tests/lib/util.js:        'host': 'localhost:2001',
node_modules/gh-tests/lib/util.js:        'host': 'localhost:2001',
node_modules/gh-tests/lib/util.js:        'host': 'localhost:2001',
node_modules/gh-tests/lib/util.js:                'host': 'localhost:2001',
node_modules/gh-users/tests/test-users.js:                                                        'host': 'localhost:2001',
```

mrvisser:
Back to @simong for some followup.

config.js
simong: I used `hostname` as that is [what node.js uses in their own documentation](https://nodejs.org/api/http.html#http_server_listen_port_hostname_backlog_callback) :(

I looked around a bit and it looks like:
 * [Apache uses `address`](http://httpd.apache.org/docs/2.2/bind.html#page-header)
 * [Nginx uses `address`](http://nginx.org/en/docs/http/ngx_http_core_module.html#listen)
 * [Elasticsearch uses `host` in their config, `address` in their documentation](http://www.elastic.co/guide/en/elasticsearch/reference/current/setup-configuration.html#settings)
 * [RabbitMQ uses `address` in their config, `interface` in their documentation](https://www.rabbitmq.com/configure.html)

What if we called it `expressListenAddress`? That way it contains:
 - `express, so you know it's about the ExpressJS http server
 - `ListenAddress`, so even if you don't know what express is, you know it's something that will be listening on a specific address.

mrvisser: Thanks, it seems from those docs that "listen" is the directive normally used. I would be fine with just "listenAddress", express is just an implementation detail.

mrvisser:
Looks good :+1:

PR #100, opened 2015-04-21
(OAuth) Client management, auth codes and token exchange

mrvisser:
Very preliminary, but functionally sanity checked. TODO:

 * Needs more tests (error scenarios, security)
 * Needs more docs


mrvisser:
Tests are basically done here. Start docs.

mrvisser:
This is ready for review. To @simong

simong:
This is prior to me reading up on OAuth / reviewing this PR so I might say dumb things.
As some of this stuff isn't entirely trivial, would you mind writing a bit of documentation as to how:
1. Application administrators can create a token for a user
2. Users can use that token to authenticate with the API

Also could you rebase this branch against master?

mrvisser:
@simong mind if I write that up in a wiki page?

mrvisser:
Assigned back to myself:

 * Need review for deprecation warnings related to `hasMany` -> `belongsToMany`
 * Needs a review of indexes. There are some unique indexes necessary to ensure token integrity

simong:
@mrvisser, Yes wiki is fine as well

mrvisser:
@simong in the meantime, this is extremely helpful:

https://aaronparecki.com/articles/2012/07/29/1/oauth2-simplified

Following the "authorization_code" / "code" grant and exchange types.

node_modules/gh-auth/lib/api.js
```diff
@@ -40,18 +47,40 @@ var ShibbolethStrategy = require('./shibboleth/strategy');
  * @param  {Object}         config              The application configuration
  */
 var init = module.exports.init = function(config) {
+    // Apply bearer token authentication
+    GrassHopper.appServer.use(function(req, res, next) {
+        if (!_hasAccessToken(req)) {
+            return next();
+        }
+
```
simong: Can we not do the following?
```javascript
passport.authenticate('bearer', {'session': false})(req, res, function() {
    if (req.ghAuthInfo && req.ghAuthInfo.user) {
        // Authentication was successful, we can safely disable CSRF for this request as we know
        // it came from a valid external site
        req._checkCSRF = false;
    }
    return next();
});
```

node_modules/gh-auth/lib/api.js
```diff
@@ -153,8 +182,113 @@ var _validateLocalUserCredentials = function(app, email, password, done) {
...
+    passport.use(new BasicStrategy({'passReqToCallback': true}, _handleOauthClientAuthentication));
+};
+
+/**
+ * This is a permutation of `_registerBasicClientPasswordAuthentication` that allows a client to
+ * provide its credentials in the request body instead of basic authentication.
+ *
+ * For clients performing POST requests, this form of authentication is acceptable since there is
+ * CSRF protection in place to ensure that the request wasn't forged. However, for GET requests, the
+ * request should use `_BasicClientPasswordAuthentication` so that credentials are trasmitted in
```
simong: `transmitted`

node_modules/gh-auth/lib/api.js
```diff
@@ -153,8 +182,113 @@ var _validateLocalUserCredentials = function(app, email, password, done) {
...
+
+    var validator = new Validator();
+    validator.check(clientId, {'code': 400, 'msg': 'A valid client id must be specified'}).isInt();
+    validator.check(clientSecret, {'code': 400, 'msg': 'A valid client secret must be specified'}).notEmpty();
+    if (validator.hasErrors()) {
+        return done(null, false, {'message': validator.getFirstError().msg});
+    }
+
+    OauthDAO.getClient(clientId, function(err, client) {
+        if (err && err.code === 500) {
```
simong: Why is there a distinction between a 500 error and a regular error?

node_modules/gh-auth/lib/api.js
```diff
@@ -289,13 +423,13 @@ var contextMiddleware = function(req, res, next) {
...
-var registerSerializeUser = function(cookieSecret) {
-    passport.serializeUser(function(authInfo, callback) {
+var _registerSerializeUser = function(cookieSecret) {
+    passport.serializeUser(function(ghAuthInfo, callback) {
         // Serialize the user id and whether the user is logged in
         // on the global admin server to the session cookie
         var sessionCookie = {
-            'userId': authInfo.user.id,
-            'isGlobalAdmin': authInfo.app.isGlobalAdmin
+            'userId': ghAuthInfo.user.id,
```
simong: When `*ClientPasswordAuthentication` is used, will this serializer run? And if it does, will `ghAuthInfo.user` be undefined?

node_modules/gh-auth/lib/oauth.js
```diff
@@ -0,0 +1,515 @@
...
+ * @param  {OauthClient}    callback.client         The updated oauth client
+ */
+var updateClient = module.exports.updateClient = function(ctx, clientId, fields, callback) {
+    clientId = GrasshopperUtil.getNumberParam(clientId);
+    fields = fields || {};
+
+    var validator = new Validator();
+    validator.check(null, {'code': 401, 'msg': 'Only administrators can update a client'}).isAdmin(ctx, ctx.app.id);
+    validator.check(clientId, {'code': 400, 'msg': 'An integer client id must be specified'}).isInt();
+
```
simong: We should probably check that at least one field is specified

node_modules/gh-auth/lib/oauth/dao.js
```diff
@@ -0,0 +1,409 @@
...
+
+        return callback(null, client);
+    });
+};
+
+/**
+ * Set the client secret of the given oauth client
+ *
+ * @param  {OauthClient}    client              The client whose secret to set
+ * @param  {Stirng}         secret              The new client secret to set
```
simong: String

node_modules/gh-core/lib/db/util.js
```diff
@@ -0,0 +1,39 @@
+
+var _ = require('lodash');
+var util = require('util');
+
+var DB = require('gh-core/lib/db');
+var log = require('gh-core/lib/logger').logger('gh-core/db/util');
+
+var findBy = module.exports.findBy = function(name, fieldName, fieldValue, opts, callback) {
```
simong: Missing jsdoc in this file

simong:
Just noticed that you mentioned you still had to self-review. I'll switch to a more higher-level approach. Apologies for the spam

node_modules/gh-auth/tests/test-oauth.js
```diff
@@ -0,0 +1,764 @@
...
+                                assert.ok(!bearerMeAfterDisable.id);
+                                return callback();
+                            });
+                        });
+                    });
+                });
+            });
+        });
+
+        it('verify listing access tokens for a user', function(callback) {
```
simong: Can you also add a test where an app admin grants a user access to a client and then verifies the user can get the access token from the API ?

node_modules/gh-auth/lib/api.js
mrvisser: No, because there is no session associated to the authentication, so a cookie will not be serialized for it.

node_modules/gh-auth/lib/api.js
mrvisser: For passport, `Internal Server Error` is the "first parameter is an err object" scenario, whereas other scenarios are an "authentication failed", in which case we use `false` as the second object.

If we put a 400 error or the like as the first parameter, passport would wrap it in a 500 which doesn't really make sense.

simong: Would you mind adding that as a comment?

mrvisser:
Just finished self-review. To @simong for review.

mrvisser:
If it helps, if you ignore whitespace on the review, it's only 4,100 lines added instead of 4,200.

simong:
Can you add a document in the wiki how this can be used by:

1. Application administrators to create a token for a user
2. Users to then use that token to authenticate with the API

node_modules/gh-auth/lib/oauth.js
```diff
@@ -0,0 +1,516 @@
...
+        return callback({'code': 401, 'msg': util.format('The specified redirect uri "%s" does not match the redirect uri configured for the client')});
+    }
+
+    AuthOauthDAO.getAuthCodeByCode(code, function(err, authCode) {
+        if (err) {
+            return callback(err);
+        } else if (authCode.OauthClientId !== client.id) {
+            return callback({'code': 401, 'msg': 'The provided authorization code does not map to the specified client'});
+        } else if (authCode.expires.getTime() <= Date.now()) {
+            // TODO, this shouldn't even work? (authCode.expires.getTime())
```
simong: Why should this not work?

node_modules/gh-auth/lib/oauth/dao.js
```diff
@@ -0,0 +1,422 @@
...
+ * Initialize the indexes and constraints of the OAuth data model
+ *
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error that occurred, if any
+ */
+var init = module.exports.init = function(callback) {
+    var sequelize = DB.getSequelize();
+
+    // We join and look up OauthClients on AppId
+    sequelize.query('CREATE INDEX oauth_clients_appid_idx ON "OauthClients" ("AppId");').complete(function(err) {
```
simong: Can these indices not be defined by using: http://sequelize.readthedocs.org/en/latest/docs/models-definition/index.html?highlight=index#indexes ?

mrvisser: I guess so, I just followed suit with what the user module was doing. But the user module is using a trigram thingy which explains that.

node_modules/gh-auth/lib/oauth.js
mrvisser: I think there was a time when it looked like those timestamps weren't going to be `Date`'s

node_modules/gh-auth/lib/oauth/rest.js
```diff
@@ -0,0 +1,197 @@
...
+    // redirect to the configured redirect URI for simplicity
+
+    // Perform the validation, authorization and grant operation
+    AuthOauth.grantAuthCode(req.ctx, req.body.client_id, req.body.for_user_id, function(err, redirectUri) {
+        if (err) {
+            return GrassHopperUtil.sendError(res, err);
+        }
+
+        // Redirect to the specified redirect uri
+        return res.redirect(redirectUri);
```
simong: This redirect might cause some issues if we're going to build a UI which does the grant + getAccessToken steps.
IIRC, you can't perform an AJAX call and NOT follow redirects.

mrvisser: Good point. We might need to extend OAuth to allow us to disable a redirect and instead return the location uri

node_modules/gh-auth/lib/oauth.js
```diff
@@ -0,0 +1,516 @@
...
+    AuthOauthDAO.getClient(clientId, function(err, client) {
+        if (err) {
+            return callback(err);
+        }
+
+        validator.check(null, {'code': 401, 'msg': 'Only administrators can disable a client'}).isAdmin(ctx, client.AppId);
+        if (validator.hasErrors()) {
+            return callback(validator.getFirstError());
+        } else if (client.disabled) {
+            return callback({'code': 400, 'msg': 'The specified client is already disabled'});
```
simong: Can we do `return callback();` here? I think it might make things easier in the UI

mrvisser: Yes that seems reasonable. I'll do the same thing w/ enable.

node_modules/gh-auth/lib/api.js
mrvisser: Comment was added

node_modules/gh-auth/lib/oauth/dao.js
mrvisser: Yup, works. That was nifty.

mrvisser:
I've followed up on review up to this point, as well as written some [API instructions](http://github.com/fronteerio/grasshopper/wiki/OAuth)

PR #105, opened 2015-05-01
Handle catch before then

mrvisser:
Something is loopy in our promise->callback wrapper. LITERALLY.

If an error is thrown in the success callback (e.g., an assert error in tests), then the `catch` handles it and re-invokes the callback.

This PR ensures that our catch handler doesn't handle callback errors, and they are instead promoted which is synonymous to our previous callback pattern.

### DIFF
--- a/node_modules/gh-core/lib/globals/sequelize.js
+++ b/node_modules/gh-core/lib/globals/sequelize.js
@@ -18,6 +18,16 @@
 
 var Promise = require('sequelize/lib/promise');
 
+/*!
+ * If an unhandled error is thrown inside the promise, we should promote it. This is needed because
+ * if an unhandled exception occurs in the `callback` during the `then` handler it is considered a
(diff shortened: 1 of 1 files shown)


node_modules/gh-core/lib/globals/sequelize.js
```diff
@@ -27,18 +37,21 @@ var Promise = require('sequelize/lib/promise');
...
             var args = Array.prototype.slice.call(arguments);
             args.unshift(null);
-            return callback.apply(null, args);
-        })
-
-        // Failure condition
-        .catch(function(err) {
-            // Send the err argument to the consumer as the first and only parameter
-            callback.call(null, err);
+            callback.apply(null, args);
```
mrvisser: Note, I took away the returns for this because `return`'s actually mean something in promises :P

PR #106, opened 2015-05-01
Keep track of whether a Sequelize promise has been dealt with

simong:


mrvisser:
I actually had this solution in place before I said "Wait a minute, I can just switch the order of the `then` and the `catch`!"

:(

PR #107, opened 2015-05-03
Import data for an organisational unit

simong:
Builds on the export branch. The idea is that this will be limited to just Timetable applications

### DIFF
--- a/node_modules/gh-orgunit/lib/internal/dao.js
+++ b/node_modules/gh-orgunit/lib/internal/dao.js
@@ -900,7 +900,7 @@ var _importOrgUnit = function(orgUnit, parent, callback) {
  * Update or create an organisational unit
  *
  * @param  {Object}     orgUnitToImport             The organisational unit to import
- * @param  {Object}     parent                      The paren organisational unit to which the organisational unit belongs
+ * @param  {Object}     parent                      The parent organisational unit to which the organisational unit belongs
  * @param  {Function}   callback                    Standard callback function
  * @param  {Object}     callback.err                An error object, if any
  * @param  {Event}      callback.persistedOrgUnit   The persisted organisational unit
@@ -946,7 +946,6 @@ var _updateOrCreateOrgUnit = function(orgUnitToImport, parent, callback) {
  * @api private
  */
 var _importOrgUnitChildren = function(children, parent, callback) {
-    children = children || [];
     if (_.isEmpty(children)) {
         return callback();
     }
@@ -957,7 +956,7 @@ var _importOrgUnitChildren = function(children, parent, callback) {
             return callback(err);
         }
 
-        _importOrgUnitChildren(children, parent, callback);
+        return _importOrgUnitChildren(children, parent, callback);
     });
 };
 
@@ -971,7 +970,6 @@ var _importOrgUnitChildren = function(children, parent, callback) {
  * @api private
  */
 var _importOrgUnitSeries = function(series, orgUnit, callback) {
-    series = series || [];
     if (_.isEmpty(series)) {
         return callback();
     }
@@ -982,7 +980,7 @@ var _importOrgUnitSeries = function(series, orgUnit, callback) {
             return callback(err);
         }
 
-        _importOrgUnitSeries(series, orgUnit, callback);
+        return _importOrgUnitSeries(series, orgUnit, callback);
     });
 };
 
@@ -1039,7 +1037,7 @@ var _updateOrCreateSeries = function(seriesToImport, orgUnit, callback) {
                 'displayName': seriesToImport.displayName,
                 'description': seriesToImport.description
             };
-            SeriesDAO.updateSerie(persistedSeries, update, callback);
+            return SeriesDAO.updateSerie(persistedSeries, update, callback);
         });
     } else {
         // Create the series
@@ -1057,7 +1055,6 @@ var _updateOrCreateSeries = function(seriesToImport, orgUnit, callback) {
  * @api private
  */
 var _importSeriesEvents = function(events, series, callback) {
-    events = events || [];
     if (_.isEmpty(events)) {
         return callback();
     }
@@ -1068,7 +1065,7 @@ var _importSeriesEvents = function(events, series, callback) {
             return callback(err);
         }
 
-        _importSeriesEvents(events, series, callback);
+        return _importSeriesEvents(events, series, callback);
     });
 };
 
@@ -1098,7 +1095,7 @@ var _importEvent = function(event, series, callback) {
                 return callback(err);
             }
 
-            // Map the user's their id to `true`
+            // Map the user's id to `true`
             var userIdOrganisers = {};
             _.each(users, function(user) {
                 userIdOrganisers[user.id] = true;
@@ -1152,13 +1149,13 @@ var _updateOrCreateEvent = function(eventToImport, series, callback) {
                 return callback(err);
             }
 
-            EventsDAO.updateEvent(persistedEvent, opts, callback);
+            return EventsDAO.updateEvent(persistedEvent, opts, callback);
         });
     } else {
         // Create the event
         opts.group = series.GroupId;
         opts.series = [series.id];
         opts.externalId = eventToImport.externalId;
-        EventsDAO.createEvent(series.AppId, opts.displayName, eventToImport.start, eventToImport.end, opts, callback);
+        return EventsDAO.createEvent(series.AppId, opts.displayName, eventToImport.start, eventToImport.end, opts, callback);
     }
 };
--- a/node_modules/gh-orgunit/lib/internal/dao.js
+++ b/node_modules/gh-orgunit/lib/internal/dao.js
@@ -25,6 +25,9 @@ var DB = require('gh-core/lib/db');
 var log = require('gh-core/lib/logger').logger('gh-orgunit');
 var UsersDAO = require('gh-users/lib/internal/dao');
 
+var EventsDAO = require('gh-events/lib/internal/dao');
+var SeriesDAO = require('gh-series/lib/internal/dao');
+
 /**
  * Create an organisational unit
  *
@@ -125,7 +128,7 @@ var deleteOrgUnit = module.exports.deleteOrgUnit = function(orgUnit, callback) {
 var getOrgUnit = module.exports.getOrgUnit = function(id, includeSeries, callback) {
     var options = {
         'where': {'id': id},
-        'include': [DB.Group]
+        'include': [DB.Group, DB.App]
     };
     if (includeSeries) {
         options.include.push(DB.Serie);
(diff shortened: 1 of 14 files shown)


simong:
Now that export has landed, this can be reviewed next

node_modules/gh-events/lib/internal/dao.js
```diff
@@ -38,14 +38,20 @@ var log = require('gh-core/lib/logger').logger('gh-events');
  * @param  {Number[]}       [opts.organiserUsers]       The id(s) of the recognised user(s) that organise the event
  * @param  {Number[]}       [opts.series]               The id(s) of the serie(s) that the event belongs to
  * @param  {String}         [opts.type]                 The type of the event
+ * @param  {String}         [opts.externalId]           The external id of the event
  */
 var createEvent = module.exports.createEvent = function(appId, displayName, start, end, opts, callback) {
+    opts = opts || {};
+    opts.organiserUsers = opts.organiserUsers || [];
+    opts.series = opts.series || [];
+
```
mrvisser: Have these been made optional? If so the docs need to change

mrvisser: Nevermind. They were already optional and the docs already reflect it

mrvisser:
@simong does this need to be rebased. There are additions of export things in here

simong:
Did another rebase

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+/* Import */
+
+/**
+ * Import a set of given organisational units, event series and events into an organisational unit
+ *
+ * @param  {MergedTreeData}     mergedTreeData      The merged tree data that contains the tree to import and the data to delete
+ * @param  {Function}           callback            Standard callback function
+ * @param  {Object}             callback.err        An error object, if any
+ */
+var importTree = module.exports.importTree = function(mergedTreeData, callback) {
```
mrvisser: What are the expected sizes of these import operations? Just asking because transactions can "overflow". I think there is usually configuration for transaction size limits on the DB side, so it's just something to keep in mind.

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+            });
+        });
+    });
+};
+
+/**
+ * Update or create an organisational unit
+ *
+ * @param  {Object}     orgUnitToImport             The organisational unit to import
+ * @param  {Object}     parent                      The paren organisational unit to which the organisational unit belongs
```
mrvisser: `parent` :/

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+    } else {
+        // Create the organisational unit
+        var opts = {
+            'description': orgUnitToImport.description,
+            'externalId': orgUnitToImport.externalId,
+            'metadata': orgUnitToImport.metadata,
+            'published': orgUnitToImport.published,
+            'parentId': parent.id
+        };
+        return createOrgUnit(orgUnitToImport.AppId, orgUnitToImport.displayName, orgUnitToImport.type, parent.GroupId, opts, callback);
```
mrvisser: I'm having trouble finding any indication that things like permissions and expected `AppId`s are being taken into consideration. For example, what is stopping someone from invoking an import on the `AppId=3` app, but specifying org units whose `AppId` field are set to `4`?

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+ * @param  {OrgUnit}    parent          The organisational unit under which the child organisational units should be imported
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error object, if any
+ * @api private
+ */
+var _importOrgUnitChildren = function(children, parent, callback) {
+    children = children || [];
+    if (_.isEmpty(children)) {
+        return callback();
+    }
```
mrvisser: If we're checking `_.isEmpty(children);`, then `children = children || [];` isn't necessary.

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+
+    var child = children.pop();
+    _importOrgUnit(child, parent, function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        _importOrgUnitChildren(children, parent, callback);
```
mrvisser: Should be `return`ed :grimacing:

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+ * Import the series in an organisational unit
+ *
+ * @param  {Object[]}   series          The series to import
+ * @param  {OrgUnit}    orgUnit         The organisational unit under which the series should be imported
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error object, if any
+ * @api private
+ */
+var _importOrgUnitSeries = function(series, orgUnit, callback) {
+    series = series || [];
```
mrvisser: This line isn't necessary

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+        return callback();
+    }
+
+    var seriesToImport = series.pop();
+    _importSeries(seriesToImport, orgUnit, function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        _importOrgUnitSeries(series, orgUnit, callback);
```
mrvisser: `return`

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+        SeriesDAO.getSerie(seriesToImport.id, false, function(err, persistedSeries) {
+            if (err) {
+                return callback(err);
+            }
+
+            var update = {
+                'displayName': seriesToImport.displayName,
+                'description': seriesToImport.description
+            };
+            SeriesDAO.updateSerie(persistedSeries, update, callback);
```
mrvisser: `return`

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+ * Import the events in a series
+ *
+ * @param  {Object[]}   events          The events to import
+ * @param  {Serie}      series          The series under which the events should be imported
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error object, if any
+ * @api private
+ */
+var _importSeriesEvents = function(events, series, callback) {
+    events = events || [];
```
mrvisser: Line note needed

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+        return callback();
+    }
+
+    var event = events.pop();
+    _importEvent(event, series, function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        _importSeriesEvents(events, series, callback);
```
mrvisser: `return`

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+        var organiserExternalIds = _.chain(event.organisers)
+            .pluck('shibbolethId')
+            .compact()
+            .value();
+        UsersDAO.getUsersByShibbolethId(series.AppId, organiserExternalIds, function(err, users) {
+            if (err) {
+                return callback(err);
+            }
+
+            // Map the user's their id to `true`
```
mrvisser: `Map the user's id to ``true```

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+            }
+
+            EventsDAO.updateEvent(persistedEvent, opts, callback);
+        });
+    } else {
+        // Create the event
+        opts.group = series.GroupId;
+        opts.series = [series.id];
+        opts.externalId = eventToImport.externalId;
+        EventsDAO.createEvent(series.AppId, opts.displayName, eventToImport.start, eventToImport.end, opts, callback);
```
mrvisser: `return`

node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -760,3 +763,402 @@ var _exportOrgUnitJSON = function(rootId, orgUnits, callback) {
...
+        'type': eventToImport.type
+    };
+    if (eventToImport.id) {
+        // Update the event
+        EventsDAO.getEvent(eventToImport.id, function(err, persistedEvent) {
+            if (err) {
+                return callback(err);
+            }
+
+            EventsDAO.updateEvent(persistedEvent, opts, callback);
```
mrvisser: `return`

node_modules/gh-orgunit/lib/internal/import.js
```diff
@@ -0,0 +1,594 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+var _ = require('lodash');
+
+var AppsConstants = require('gh-apps/lib/constants');
+var GrasshopperUtil = require('gh-core/lib/util');
```
mrvisser: I've seen a mix of `GrassHopperUtil` and `GrasshopperUtil` nomenclature. I agree with this one, so lets try and standardize on this moving forward and eventually submit a sweeping PR

node_modules/gh-orgunit/lib/internal/import.js
```diff
@@ -0,0 +1,594 @@
...
+var log = require('gh-core/lib/logger').logger('gh-orgunit');
+var util = require('util');
```
mrvisser: Should be under `_` import.

node_modules/gh-orgunit/lib/internal/dao.js
mrvisser: I suppose this is the DAO layer, and so there is probably validation for this on the API layer (haven't got there yet).

mrvisser: FWIW, I think the fact that we look at an AppId in the data rather than hand it into the DAO as a DB object tripped me up on the whole validation thing. Preferrably we could hand the `app` as the first parameter of the import method(s), but if we need to support multi-app imports then maybe that isn't feasible.

node_modules/gh-orgunit/lib/internal/import.js
```diff
@@ -0,0 +1,594 @@
...
+
+                    // Import the merged data and remove the data that needs to be removed (if any)
+                    return OrgUnitDAO.importTree(mergedData, callback);
+                });
+            });
+        });
+    });
+};
+
+var getParentIfNecessary = function(orgUnit, callback) {
```
mrvisser: Needs docs. Also prefix w/ `_`.

FWIW, when OAuth lands this can also be `_.ghIf(orgUnit.parentId, _getParent, orgUnit, function(err, parent) { ...`

:trollface:

node_modules/gh-orgunit/lib/internal/import.js
```diff
@@ -0,0 +1,594 @@
...
+    seenTypes[orgUnit.type] = currentLevel;
+
+    if (orgUnit.description) {
+        validator.check(orgUnit.description, {'code': 400, 'msg': 'A description can be at most 1000 characters long'}).isMediumString();
+    }
+    if (orgUnit.metadata) {
+        if (_.isString(orgUnit.metadata)) {
+            try {
+                orgUnit.metadata = JSON.parse(orgUnit.metadata);
+            } catch (e) {
```
mrvisser: We should keep some indication at least for support purposes of what went wrong (stack trace). Probably even pass the `e.message` into the result, especially if it gives more information about where parsing failed.

node_modules/gh-orgunit/lib/rest.js
```diff
@@ -578,3 +581,43 @@ var exportOrgUnit = function(req, res) {
...
+ * @Path        /orgunit/{id}/import
+ * @PathParam   {number}            id              The id of the organisational unit into which to import the data
+ * @FormParam   {string}            data            The organisational units, event series and events to import
+ * @FormParam   {boolean}           deleteMissing   Whether to delete items from the database that aren't in the import data
+ */
+var importOrgUnit = function(req, res) {
+    OrgUnitImport.importOrgUnit(req.ctx, req.params.id, req.body.data, req.body.deleteMissing, function(errors) {
+        if (errors) {
+            // Because of the large potential of validation errors, the import API returns everything
+            // that is wrong with the provided data. We provide all the information to the user
```
mrvisser: `We provide all the information to the user` seems out of place

node_modules/gh-orgunit/lib/rest.js
```diff
@@ -578,3 +581,43 @@ var exportOrgUnit = function(req, res) {
...
+            if (_.isArray(errors)) {
+                var body = _.map(errors, function(error) {
```
mrvisser: This `map` can be a `pluck`

node_modules/gh-orgunit/lib/rest.js
```diff
@@ -578,3 +581,43 @@ var exportOrgUnit = function(req, res) {
...
+                    return error.msg;
+                }).join('\n');
+                var code = 400;
+                if (errors.length === 1) {
+                    code = errors[0].code;
+                }
+                return res.status(code).send(body);
```
mrvisser: I think for the whole `errors.length === 1` case, we could just take `errors[0]` and do what happens in the `else` case.

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -0,0 +1,806 @@
+/**
+ * Copyright (c) 2014 "Fronteer LTD"
```
mrvisser: `2015`

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -0,0 +1,806 @@
...
+                            importData.children[0].series[0].externalId = importData.children[0].series[0].events[0].externalId;
+                            OrgUnitTestsUtil.assertImportOrgUnitImportFails(global.tests.admins.cam2013.client, part.id, importData, false, ['Duplicate externalId'], function() {
+
+                                // Import the structure into the part
+                                importData = _getTestTree('1part-2module-1series-1event');
+                                OrgUnitTestsUtil.assertImportOrgUnit(global.tests.admins.cam2013.client, part.id, importData, false, function() {
+
+                                    // Verify we can import the SAME structure into ANOTHER part. This is necessary
+                                    // as multiple consumers can import data, but they don't necessarily know about
+                                    // each other's external ids.
```
mrvisser: Does this mean that we can't add a DB constraint to make `externalId`'s globally unique in the system?

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -0,0 +1,806 @@
...
+
+                // Too long displayName
+                var importData = _getTestTree('1part-2module-1series-1event');
+                importData.children[0].displayName = TestsUtil.generateString(257);
+                OrgUnitTestsUtil.assertImportOrgUnitImportFails(global.tests.admins.cam2013.client, part.id, importData, false, ['displayName'], function() {
+
+                    // Too long type
+                    importData = _getTestTree('1part-2module-1series-1event');
+                    importData.children[0].type = TestsUtil.generateString(33);
+                    OrgUnitTestsUtil.assertImportOrgUnitImportFails(global.tests.admins.cam2013.client, part.id, importData, false, ['type'], function() {
```
mrvisser: Aren't there an enumerated set of types we can validate against? Or can consumers effectively make their own types?

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -0,0 +1,806 @@
...
+         */
+        it('verify basic events validation', function(callback) {
+            OrgUnitTestsUtil.assertCreateOrgUnit(global.tests.admins.cam2013.client, 'Test part', 'part', {}, function(part) {
+
+                // Too long displayName
+                var importData = _getTestTree('1part-1module-1series-1event');
+                importData.children[0].series[0].events[0].displayName = TestsUtil.generateString(257);
+                OrgUnitTestsUtil.assertImportOrgUnitImportFails(global.tests.admins.cam2013.client, part.id, importData, false, ['displayName'], function() {
+
+                    // Mising start date
```
mrvisser: typo

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -0,0 +1,806 @@
...
+         */
+        it('verify deleting missing data', function(callback) {
+            // Set up an organisational unit with some structure underneath it
+            OrgUnitTestsUtil.assertCreateOrgUnit(global.tests.admins.cam2013.client, 'Test part', 'part', {}, function(part) {
+                var importData = _getTestTree('1part-2module-2series-2event');
+                OrgUnitTestsUtil.assertImportOrgUnit(global.tests.admins.cam2013.client, part.id, importData, false, function() {
+                    OrgUnitTestsUtil.assertExportOrgUnit(global.tests.admins.cam2013.client, part.id, null, function(dataPostFirstImport) {
+
+                        // Subscribe a user to a few series. This allows us to verify that series get
+                        // removed from users their calenders when they get removed through an import
```
mrvisser: `calendars`

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -0,0 +1,806 @@
...
+            });
+        });
+    });
+
+    describe('DAO', function() {
+
+        /**
+         * Test that verifies that the import logic uses transactions
+         */
+        it.skip('verify transactions', function(callback) {
```
mrvisser: Is this test being skipped?

node_modules/gh-orgunit/tests/util.js
```diff
@@ -1124,6 +1124,72 @@ var _getAllEvents = function(client, id, callback) {
...
+ * @param  {Number}             id                  The id of the organisational unit into which to import the data
+ * @param  {Object}             data                The organisational units, event series and events to import
+ * @param  {Boolean}            deleteMissing       Whether to delete items from the database that aren't in the import data
+ * @param  {Number}             code                The expected HTTP error code
+ * @param  {Function}           callback            Standard callback function
+ */
+var assertImportOrgUnitFails = module.exports.assertImportOrgUnitFails = function(client, id, data, deleteMissing, code, callback) {
+    client.orgUnit.importOrgUnit(id, data, deleteMissing, function(err) {
+        assert.ok(err);
+        assert.strictEqual(err.code, code, JSON.stringify(err));
```
mrvisser: There are a couple `JSON.stringify`'s in here that can go

node_modules/gh-rest/lib/rest/orgunit.js
```diff
@@ -293,4 +293,25 @@ module.exports = function(client) {
...
+     * @param  {Object}         callback.body                   The JSON response from the REST API
+     * @param  {Response}       callback.response               The response object as returned by requestjs
+     * @see gh-orgunit/lib/rest.js for more information
+     */
+    client.orgUnit.importOrgUnit = function(id, importData, deleteMissing, callback) {
+        var url = '/api/orgunit/' + client.util.encodeURIComponent(id) + '/import';
+        var data = {
+            'data': JSON.stringify(importData),
+            'deleteMissing': deleteMissing
+        };
```
mrvisser: We could also consider accepting `Content-Type: application/json` with this data which might be more straightforward and web-like. But I don't think this is a big issue.

mrvisser:
I'd suggest adding a unique index on the `externalId`s if possible to do so.

mrvisser:
I've finished review, this looks really good. Just a few relatively small suggestions and questions, back to @simong for follow-up.

node_modules/gh-orgunit/tests/test-import.js
simong: Correct. Departments might have clashing external ids with each other. We only require an external ID to be unique within the organisational unit structure they are importing. This could cause an issue if they start importing on part basis and later move to importing entire courses, but that's something we've agreed upon with Cambridge.

node_modules/gh-orgunit/tests/test-import.js
simong: The import API is scoped to Timetable types. This means only `course`, `subject`, `part` and `module` are valid types. The `verify only timetable types can be imported` test verifies this

node_modules/gh-orgunit/tests/test-import.js
simong: Yes, my bad, I thought I had finished this. The transaction logic requires transaction support in the EventsDAO, SeriesDAO and OrgUnitDAO and I think I wanted to outscope that from this PR.

node_modules/gh-orgunit/lib/internal/dao.js
simong: The import file and rollover file work on the highest (course) level. At a guess, I think the largest course in Cambridge contains maybe 20k events, 2k series and about 500 org units. 

Postgres has a hard limit of 2^31 number of queries that can be ran, we're probably not going to reach this one.
When a transaction gets committed, postgres needs to be able to process them of course, I assume this is a matter of disk size and available IO. I'll do a couple of imports against the RDS database, but I suspect our datasize is small enough to not pose a problem

mrvisser:
There is absolutely no chance we'll hit a `2^31` query limit in a transaction. Also, I think it you're correct that it's the only limitation on transaction size on Postgres, so I'm satisfied without any further testing :)

mrvisser:
Followup looks good :+1:

PR #110, opened 2015-05-19
Always run in UTC + iCal name

simong:
Couple of minor fixes

### DIFF
--- a/app.js
+++ b/app.js
@@ -18,6 +18,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// Always run in UTC, no matter what
+process.env.TZ = 'UTC';
+
 var yargs = require('yargs');
 
 var GrassHopper = require('gh-core/lib/api');
--- a/node_modules/gh-tests/lib/util.js
+++ b/node_modules/gh-tests/lib/util.js
@@ -587,11 +587,13 @@ var parseRssCalendar = module.exports.parseRssCalendar = function(calendar, call
  * Parse an iCal calendar
  *
  * @param  {String}             calendar                The iCal calendar
+ * @param  {String}             appName                 The name of the application from which the calendar is retrieved
+ * @param  {String}             calendarName            The name of the user or object that owns the calendar
  * @param  {Function}           callback                Standard callback function
  * @param  {Object}             callback.calendar       The parsed iCal calendar
  * @throws {AssertionError}                             Error thrown when an assertion failed
  */
-var parseIcalCalendar = module.exports.parseIcalCalendar = function(calendar, callback) {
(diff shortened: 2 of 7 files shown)


app.js
```diff
@@ -18,6 +18,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// Always run in UTC, no matter what
+process.env.TZ = 'UTC';
```
nicolaasmatthijs: Is this required? Wouldn't we always be working with UTC timestamp when using date objects and "getTime()"?

node_modules/gh-tests/lib/util.js
```diff
@@ -596,10 +596,14 @@ var parseIcalCalendar = module.exports.parseIcalCalendar = function(calendar, ca
...
-        // assert.ok(calendar.fields);
-        // assert.ok(calendar.fields.PRODID);
+        // Assert we return a name for this calendar
+        var prodId = _.find(calendar.rawFields, {'key': 'PRODID'});
+        assert.ok(prodId);
+        assert.ok(prodId.value);
+
+        var calName = _.find(calendar.rawFields, {'key': 'X-WR-CALNAME'});
+        assert.ok(calName);
+        assert.ok(calName.value);
```
nicolaasmatthijs: Do we have enough information to assert that the name is the expected one?

nicolaasmatthijs:

app.js
simong: This is only an issue when your machine isn't running in UTC (fwiw, all our servers are).
New date objects default to the `TZ` timezone when created. It becomes a real issue when serialising a calendar to ical (the cozy-ical library doesn't seem to handle this terribly well).

(fwiw, we had to do it in OAE as well: https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-util/lib/oae.js#L62)

simong:
followed up, back to @nicolaasmatthijs

nicolaasmatthijs:
Left one final question.

It also looks like there's a test failure, although it's probably not related to this PR:

```
1) Oauth authentication DAO verify two access tokens with same token value cannot be created:
      Uncaught AssertionError: (key: "updatedAt") false == true
      + expected - actual
      -false
      +true
      
  
  2) Oauth authentication Access Tokens verify exchange authorization code:
      Uncaught AssertionError: (key: "updatedAt") false == true
      + expected - actual
      -false
      +true
```

nicolaasmatthijs:
Back to @simong

simong:
Those OAuth tests have been resolved, assigning to @mrvisser for review

mrvisser:
@simong there are still a bunch of scripts that don't set UTC at the top (in `etc/scripts...`). Shouldn't they?

mrvisser:
Back to @simong for feedback

simong:
I'll address that in the rollover branch, back to @mrvisser

PR #111, opened 2015-05-20
Increase accuracy of year rollover date calculations

ucamhal:
This gives grasshopper the same rollover behaviour as Timetable 3.

The command which generates the dataset used in the tests is available here: https://github.com/CUL-DigitalServices/timetable-3/pull/1


simong:
Thanks a bunch Hal, I've tweaked some things so it matches our styleguide and merged this into my rollover branch

PR #112, opened 2015-05-26
Fix a couple failing tests

mrvisser:
This fixes a consistent failure that currently exists in export, as well as an intermittent failure I came across in the meantime.

### DIFF
--- a/node_modules/gh-auth/tests/util.js
+++ b/node_modules/gh-auth/tests/util.js
@@ -341,12 +341,14 @@ var assertOauthClient = module.exports.assertOauthClient = TestsUtil.createAsser
                 'redirectUri': _.isString,
                 'disabled': function(actualDisabled) {
                     if (_.isString(actualDisabled)) {
+                        // TODO: Enable this when https://github.com/sequelize/sequelize/pull/3733 is resolved
                         // It must have been disabled sometime after it was created, or before or equal
                         // to the last update time
-                        TestsUtil.assertTimestamp(actualDisabled, {
-                            'afterOrEqual': actualOauthClient.createdAt,
-                            'beforeOrEqual': actualOauthClient.updatedAt
-                        });
+                        // TestsUtil.assertTimestamp(actualDisabled, {
+                        //     'afterOrEqual': actualOauthClient.createdAt,
+                        //     'beforeOrEqual': actualOauthClient.updatedAt
+                        // });
+                        return true;
                     } else if (_.isNull(actualDisabled)) {
                         return true;
                     } else {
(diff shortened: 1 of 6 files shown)


mrvisser:
There still appear to be issues in OAuth with respect to the updated timestamp, but I can't isolate a case of it still.

node_modules/gh-series/tests/util.js
```diff
@@ -577,14 +577,13 @@ var assertSubscribeSeries = module.exports.assertSubscribeSeries = function(clie
...
             assertGetSeriesEvents(client, id, null, null, null, null, function(events) {
-
                 // Generate a broad window for the user's calendar
                 var start = moment().subtract(30, 'days').format();
                 var end = moment().add(30, 'days').format();
 
                 // If the serie contained events, we generate a window around the serie events
                 // and assert that the events are returned in the user's calendar
-                if (!_.isEmpty(events.results.length)) {
+                if (!_.isEmpty(events.results)) {
```
simong: I fixed this in one of my branches as well and IIRC there's another spot in this file where we do `isEmpty(... .length)`

mrvisser: You're right. I'll fix that and I found another spot with bogus start/end handling, so I'll push a commit for that as well.

simong:
The OAuth issue is because to the updatedAt and createdAt values use a different date object when creating the record. This can cause the updatedAt value be 1 millisecond earlier than the createdAt value. There's an [open PR for Sequelize that addresses this](https://github.com/sequelize/sequelize/pull/3733/files).

node_modules/gh-series/tests/util.js
```diff
@@ -577,16 +577,17 @@ var assertSubscribeSeries = module.exports.assertSubscribeSeries = function(clie
...
             assertGetSeriesEvents(client, id, null, null, null, null, function(events) {
-
                 // Generate a broad window for the user's calendar
                 var start = moment().subtract(30, 'days').format();
                 var end = moment().add(30, 'days').format();
 
-                // If the serie contained events, we generate a window around the serie events
+                // If the series contained events, we generate a window around the series events
                 // and assert that the events are returned in the user's calendar
-                if (!_.isEmpty(events.results.length)) {
```
simong: There is a similar bug on line 650 in this file

mrvisser:
Followed up, and fixed a couple more failures. Back to @simong

node_modules/gh-auth/tests/util.js
```diff
@@ -882,11 +886,13 @@ var assertOauthAccessToken = module.exports.assertOauthAccessToken = TestsUtil.c
                     });
                 },
                 'updatedAt': function(actualUpdatedAt) {
+                    // TODO: Enable this when https://github.com/sequelize/sequelize/pull/3733 is resolved
```
simong: It seems like the sequelize guys moved faster than we did. Sequelize 3.1.1 has been released and contains that PR.

simong:
PR looks good, feel free to merge into master when undoing those TODOs

PR #114, opened 2015-05-30
Change password

simong:
Adds support for changing a local user's and a global administrator's password

### DIFF
--- a/node_modules/gh-admins/lib/api.js
+++ b/node_modules/gh-admins/lib/api.js
@@ -16,6 +16,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+var AuthUtil = require('gh-auth/lib/util');
 var GrasshopperUtil = require('gh-core/lib/util');
 var log = require('gh-core/lib/logger').logger('gh-users');
 var Validator = require('gh-core/lib/validator').Validator;
@@ -156,6 +157,58 @@ var updateGlobalAdmin = module.exports.updateGlobalAdmin = function(ctx, id, dis
         }
 
         return AdminsDAO.updateGlobalAdmin(globalAdmin, displayName, callback);
+    });
+};
+
+/**
+ * Change a global administrator's password. The administator making the change will have to provide
+ * their own password. This is done to thwart situations where an administrator's session gets hijacked
+ *
+ * @param  {Context}        ctx                     Standard context containing the current user and the current app
+ * @param  {String}         id                      The id of the global administrator for which to change the password
+ * @param  {String}         newPassword             The new password for the global administrator
+ * @param  {String}         password                The password of the global administrator making the change
+ * @param  {Function}       callback                Standard callback function
+ * @param  {Object}         callback.err            An error that occurred, if any
+ */
+var changePassword = module.exports.changePassword = function(ctx, id, newPassword, password, callback) {
+    // Ensure the global administrator id is a valid number
+    id = GrasshopperUtil.getNumberParam(id);
+
+    // Validation
+    var validator = new Validator();
+    validator.check(null, {'code': 401, 'msg': 'Only global administrators can change the password of a global administrator'}).isGlobalAdmin(ctx);
+    validator.check(id, {'code': 400, 'msg': 'A valid global administrator id must be provided'}).isInt();
+    validator.check(newPassword, {'code': 400, 'msg': 'A valid new password must be provided'}).notEmpty();
+    validator.check(newPassword, {'code': 400, 'msg': 'A valid new password that is 6 characters or longer must be provided'}).len(6);
+    validator.check(password, {'code': 400, 'msg': 'A valid password must be provided'}).notEmpty();
+    if (validator.hasErrors()) {
+        return callback(validator.getFirstError());
+    }
+
+    // Ensure the global administrator that is making the change specified their own password
+    if (!AuthUtil.comparePassword(password, ctx.user.password)) {
+        return callback({'code': 401, 'msg': 'An incorrect password was provided'});
+    }
+
+    // Verify that the provided global administrator exists
+    getGlobalAdmin(ctx, id, function(err, globalAdmin) {
+        if (err) {
+            return callback(err);
+        }
+
+        AdminsDAO.changePassword(globalAdmin, newPassword, function(err) {
+            if (err) {
+                return callback(err);
+            }
+
+            log().info({
+                'actor': ctx.user,
+                'globalAdmin': globalAdmin
+            }, 'Updated the password of a global administrator');
+
+            return callback();
+        });
     });
 };
 
--- a/node_modules/gh-admins/tests/test-admins.js
+++ b/node_modules/gh-admins/tests/test-admins.js
@@ -310,4 +310,93 @@ describe('Admins', function() {
             });
         });
     });
+
+    describe('Changing a global admin\'s password', function() {
+
+        /**
+         * Test that verifies that a global administrator's password can be updated
+         */
+        it('verify a global administrator\'s password can be updated', function(callback) {
+            var username = TestsUtil.generateTestUserId();
+            AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, 'password', 'display name', function(createdGlobalAdmin, createdGlobalAdminClient) {
+
+                AdminsTestsUtil.assertChangePassword(globalAdminClient, createdGlobalAdmin.id, 'a new password', globalAdminClient.options.password, function(newClient) {
+                    return callback();
+                });
+            });
+        });
+
+        /**
+         * Test that verifies validation when changing a global administrator's password
+         */
+        it('verify validation when changing a global administrator\'s password', function(callback) {
+            var username = TestsUtil.generateTestUserId();
+            AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, 'password', 'display name', function(createdGlobalAdmin, createdGlobalAdminClient) {
+
+                var newPassword = 'new password';
+
+                // Invalid global administrator id
+                AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, 'Not a number', newPassword, globalAdminClient.options.password, 400, function() {
+                    // Unknown global administrator id
+                    AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, -1, newPassword, globalAdminClient.options.password, 404, function() {
+                        AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, 123123123, newPassword, globalAdminClient.options.password, 404, function() {
+
+                            // Missing new password
+                            AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, createdGlobalAdmin.id, null, globalAdminClient.options.password, 400, function() {
+
+                                // Too short new password
+                                AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, createdGlobalAdmin.id, 's', globalAdminClient.options.password, 400, function() {
+
+                                    // Missing password of the global administrator making the change
+                                    AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, createdGlobalAdmin.id, newPassword, null, 400, function() {
+                                        return callback();
+                                    });
+                                });
+                            });
+                        });
+                    });
+                });
+            });
+        });
+
+        /**
+         * Test that verifies authorization when changing a global administrator's password
+         */
+        it('verify authorization when changing a global administrator\'s password', function(callback) {
+            var username = TestsUtil.generateTestUserId();
+            AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, 'password', 'display name', function(createdGlobalAdmin, createdGlobalAdminClient) {
+
+                var newPassword = 'new password';
+
+                // Anonymous users can't change other global administrators their password
+                TestsUtil.getAnonymousAppUserClient(global.tests.apps.cam2014, function(anonymousClient) {
+                    AdminsTestsUtil.assertChangePasswordFails(anonymousClient, createdGlobalAdmin.id, newPassword, globalAdminClient.options.password, 404, function() {
+                        TestsUtil.getAnonymousGlobalAdminRestClient(function(anonymousGlobalAdminClient) {
+                            AdminsTestsUtil.assertChangePasswordFails(anonymousGlobalAdminClient, createdGlobalAdmin.id, newPassword, globalAdminClient.options.password, 401, function() {
+
+                                // Regular users can't change global administrators their password
+                                TestsUtil.generateTestUsers(global.tests.apps.cam2013, 1, false, function(simon) {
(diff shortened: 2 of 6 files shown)


simong:
Assigning to @mrvisser for review

node_modules/gh-admins/tests/test-admins.js
```diff
@@ -310,4 +310,85 @@ describe('Admins', function() {
...
+         * Test that verifies that a global administrator's password can be updated
+         */
+        it('verify a global administrator\'s password can be updated', function(callback) {
+            var username = TestsUtil.generateTestUserId();
+            AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, 'password', 'display name', function(createdGlobalAdmin, createdGlobalAdminClient) {
+
+                AdminsTestsUtil.assertChangePassword(globalAdminClient, createdGlobalAdmin.id, 'a new password', function(newClient) {
+                    return callback();
+                });
+            });
```
mrvisser: Should we verify a global admin can change another global admin's password?

node_modules/gh-admins/lib/api.js
```diff
@@ -160,6 +160,39 @@ var updateGlobalAdmin = module.exports.updateGlobalAdmin = function(ctx, id, dis
...
+    }
+
+    // Verify that the provided global administrator exists
+    getGlobalAdmin(ctx, id, function(err, globalAdmin) {
+        if (err) {
+            return callback(err);
+        }
+
+        return AdminsDAO.changePassword(globalAdmin, newPassword, callback);
+    });
```
mrvisser: I feel like this kind of endpoint should also take the current password of the global admin in context, just to ensure this operation is not available for a global admin whose session was hijacked somehow (e.g., some CORS issue, XSS, etc...)

node_modules/gh-admins/tests/test-admins.js
```diff
@@ -310,4 +310,85 @@ describe('Admins', function() {
...
+        it('verify validation when changing a global administrator\'s password', function(callback) {
+            var username = TestsUtil.generateTestUserId();
+            AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, 'password', 'display name', function(createdGlobalAdmin, createdGlobalAdminClient) {
+
+                var newPassword = 'new password';
+
+                // Invalid global administrator id
+                AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, 'Not a number', newPassword, 400, function() {
+                    // Unknown global administrator id
+                    AdminsTestsUtil.assertChangePasswordFails(globalAdminClient, -1, newPassword, 404, function() {
```
mrvisser: Can we also try an id that *could* exist but doesn't? Like `2342342342`

node_modules/gh-admins/tests/test-admins.js
mrvisser: I see this is already being tested in the validation/authorization endpoint

node_modules/gh-admins/tests/util.js
```diff
@@ -270,6 +270,66 @@ var assertUpdateGlobalAdminFails = module.exports.assertUpdateGlobalAdminFails =
...
+ */
+var assertChangePassword = module.exports.assertChangePassword = function(client, id, newPassword, callback) {
+    // Change the global administrator's password
+    client.admin.changePassword(id, newPassword, function(err) {
+        assert.ok(!err);
+
+        // Grab the admin's profile information so we can attempt a login
+        // and verify the password was changed successfully
+        assertGetAllGlobalAdmins(client, null, function(globalAdmins) {
+            var globalAdmin = _.findWhere(globalAdmins, {'id': id});
```
mrvisser: Can you create an `assertGetGlobalAdmin` method that covers getting all and finding? I'm guessing it should return a `false`y value instead of an error code if the specified admin doesn't exist, just due to the semantics.

node_modules/gh-rest/lib/rest/admin.js
```diff
@@ -81,6 +81,22 @@ module.exports = function(client) {
...
     /**
+     * Change a global administrator's password
+     *
+     * @param  {Number}             id                          The id of the global administrator for which to change the password
+     * @param  {String}             newPassword                 The new password for the global administrator
+     * @param  {Function}           callback                    Standard callback function
+     * @param  {Object}             callback.err                An error that occurred, if any
+     */
+    client.admin.changePassword = function(id, newPassword, callback) {
+        var url = '/api/admins/' + client.util.encodeURIComponent(id) + '/password';
```
mrvisser: It would be nice to have something like a `uprintf` (the `u` is for "url") that is similar to `sprintf`/`util.format`, except it will run `encodeURIComponent` on all the parameters:

`var url = uprintf('/api/admins/%s/password', id);`

node_modules/gh-users/lib/api.js
```diff
@@ -295,6 +296,64 @@ var updateUser = module.exports.updateUser = function(ctx, userId, update, callb
 };
 
 /**
+ * Change a user's local password
+ *
+ * @param  {Context}        ctx                     Standard context containing the current user and the current app
+ * @param  {string}         userId                  The id of user for which to change the local password
+ * @param  {string}         newPassword             The new password for the user
+ * @param  {string}         [oldPassword]           The previous password for the user. This is only required when the current user is not an administrator
```
mrvisser: I believe these should be `String` instead of `string`. Though I have been leaning more lately towards using JSDoc (syntax is much better documented) which uses lower-case for native types.

node_modules/gh-users/lib/api.js
```diff
@@ -295,6 +296,64 @@ var updateUser = module.exports.updateUser = function(ctx, userId, update, callb
...
+    }
+
+    // Ensure the user exists
+    UsersDAO.getUser(userId, function(err, user) {
+        if (err) {
+            return callback(err);
+        }
+
+        // Ensure the user is using local authentication
+        if (!user.password) {
```
mrvisser: Can we join this as an `else if` to the above?

node_modules/gh-users/tests/test-users.js
```diff
@@ -522,4 +523,137 @@ describe('Users', function() {
...
+        it('verify validation when changing a user\'s password', function(callback) {
+            TestsUtil.generateTestUsers(global.tests.apps.cam2014, 1, false, function(simon) {
+
+                var newPassword = 'new password';
+                var oldPassword = simon.client.options.password;
+
+                // Invalid user id
+                UsersTestsUtil.assertChangePasswordFails(simon.client, 'Not a number', newPassword, oldPassword, 400, function() {
+                    // Unknown user id
+                    UsersTestsUtil.assertChangePasswordFails(simon.client, -1, newPassword, oldPassword, 404, function() {
```
mrvisser: Test with something like `2342342423423`

node_modules/gh-users/lib/api.js
```diff
@@ -295,6 +296,64 @@ var updateUser = module.exports.updateUser = function(ctx, userId, update, callb
...
+            return callback({'code': 400, 'msg': 'This user is not using local authentication'});
+
+        // Regular users have to provide an old password
+        } else if (!ctx.user.canAdmin(user.AppId)) {
+            validator.check(oldPassword, {'code': 400, 'msg': 'A valid old password must be provided'}).notEmpty();
+            validator.check(oldPassword, {'code': 400, 'msg': 'A valid old password must be provided'}).len(6);
```
mrvisser: I don't think we need to do the too short check for the old password.

node_modules/gh-users/tests/test-users.js
```diff
@@ -522,4 +523,137 @@ describe('Users', function() {
...
+                        UsersTestsUtil.assertChangePasswordFails(simon.client, simon.profile.id, null, oldPassword, 400, function() {
+
+                            // Too short new password
+                            UsersTestsUtil.assertChangePasswordFails(simon.client, simon.profile.id, 's', oldPassword, 400, function() {
+
+                                // Missing old password
+                                UsersTestsUtil.assertChangePasswordFails(simon.client, simon.profile.id, newPassword, null, 400, function() {
+
+                                    // Too short old password
+                                    UsersTestsUtil.assertChangePasswordFails(simon.client, simon.profile.id, newPassword, 's', 400, function() {
```
mrvisser: I think we can get rid of this check

node_modules/gh-users/tests/test-users.js
```diff
@@ -522,4 +523,137 @@ describe('Users', function() {
...
+
+                                        // Verify administrators don't need to provide an old password
+                                        UsersTestsUtil.assertChangePassword(global.tests.admins.cam2014.client, simon.profile.id, 'a new password', null, function(newClient) {
```
mrvisser: Perhaps administrators should provide *their own* old password for the same reason mentioned before.

mrvisser: Another potential issue is that compromised OAuth clients would be able to hijack accounts.

node_modules/gh-admins/lib/api.js
mrvisser: Also we should have an `info` log

node_modules/gh-users/lib/api.js
```diff
@@ -295,6 +296,64 @@ var updateUser = module.exports.updateUser = function(ctx, userId, update, callb
...
+                return callback(err);
+            } else if (!canChangePassword) {
+                log().warn({'id': userId, 'actor': ctx.user.id}, 'Unauthorized attempt at changing a user\'s password');
+                return callback({'code': 401, 'msg': 'You are not allowed to update this user\'s password'});
+            }
+
+            var update = {
+                'password': AuthUtil.hashPassword(newPassword)
+            };
+            UsersDAO.updateUser(user, update, callback);
```
mrvisser: Should have an `info` log that a user's password has changed, including the user who changed it

node_modules/gh-users/tests/test-users.js
```diff
@@ -522,4 +523,137 @@ describe('Users', function() {
...
+        /**
+         * Test that verifies authorization when changing a user's password
+         */
+        it('verify authorization when changing a user\'s password', function(callback) {
+            TestsUtil.generateTestUsers(global.tests.apps.cam2014, 2, false, function(simon, nico) {
+
+                var newPassword = 'new password';
+                var oldPassword = simon.client.options.password;
+
+                // Anonymous users can't change other users their password
```
mrvisser: I don't think it's super important to change this because I know exactly what it's saying, but I think I'll keep nagging you on this grammar structure any way :)

`// Anonymous users can't change other user passwords`

:sunglasses:

node_modules/gh-users/tests/test-users.js
```diff
@@ -522,4 +523,137 @@ describe('Users', function() {
...
+                                    });
+                                });
+                            });
+                        });
+                    });
+                });
+            });
+        });
+
+        it('verify only users with a local password can be changed', function(callback) {
```
mrvisser: No doc on this test? I'd have expected more from an Apereo fellow.

node_modules/gh-users/tests/util.js
```diff
@@ -297,6 +298,73 @@ var assertUpdateAdminStatusFails = module.exports.assertUpdateAdminStatusFails =
...
+    client.user.changePassword(id, newPassword, oldPassword, function(err) {
+        assert.ok(!err);
+
+        // Grab the user's profile information so we can attempt a local login
+        // and verify the password was changed successfully
+        assertGetUser(client, id, function(user) {
+
+            // Get the object of the application the user belongs to so we can create a REST client
+            TestsUtil.getGlobalAdminRestClient(function(globalAdminClient) {
+                AppTestsUtil.assertGetApp(globalAdminClient, user.AppId, null, function(app) {
```
mrvisser: http://www.imdb.com/title/tt0120772/

I think we can just say `Get the application the user belongs to...`

mrvisser:
Overall looks good. Just a couple security probes and minor style issues. Back to @simong

node_modules/gh-admins/tests/util.js
simong: I've created the function, but it will throw if the admin can't be found. That seems to be more inline with our other `assertGetFoo` utilities

simong:
I took out the users password bits and followed up. Unfortunately, I wasn't really paying attention when rebasing and squashed the follow-up :(

node_modules/gh-admins/lib/api.js
```diff
@@ -160,6 +161,58 @@ var updateGlobalAdmin = module.exports.updateGlobalAdmin = function(ctx, id, dis
...
+
+        AdminsDAO.changePassword(globalAdmin, newPassword, function(err) {
+            if (err) {
+                return callback(err);
+            }
+
+            log().info({
+                'actor': ctx.user,
+                'globalAdmin': globalAdmin
+            }, 'Updated the password of a global administrator');
```
mrvisser: Is this logging the password hash? If so, we shouldn't

node_modules/gh-admins/tests/test-admins.js
```diff
@@ -310,4 +310,93 @@ describe('Admins', function() {
...
+        /**
+         * Test that verifies authorization when changing a global administrator's password
+         */
+        it('verify authorization when changing a global administrator\'s password', function(callback) {
+            var username = TestsUtil.generateTestUserId();
+            AdminsTestsUtil.assertCreateGlobalAdmin(globalAdminClient, username, 'password', 'display name', function(createdGlobalAdmin, createdGlobalAdminClient) {
+
+                var newPassword = 'new password';
+
+                // Anonymous users can't change other global administrators their password
```
mrvisser: `... can't change global administrator passwords`

node_modules/gh-admins/tests/util.js
```diff
@@ -61,6 +61,32 @@ var assertMe = module.exports.assertMe = TestsUtil.createAssertionFunction({
 });
 
 /**
+ * Get a global administrator by their id
+ *
+ * @param  {RestClient}     client                  The REST client to make the request with
+ * @param  {Object}         id                      The id of the global administrator to retrieve
+ * @param  {Function}       callback                Invoked when the global administrator has been retrieved
+ * @param  {GlobalAdmin}    callback.globalAdmin    The global administrator
+ * @throws {AssertionError}                             Thrown if any request or assertions fail
```
mrvisser: Spacing

mrvisser:
Looks good. Just a probe about potentially logging password hashes and other really minor things. Feel free to merge once you've addressed / given feedback.

node_modules/gh-admins/lib/api.js
simong: It isn't, when we serialize a `GlobalAdmin` or `User` instance to JSON we delete the `password` property from the object

simong: See https://github.com/fronteerio/grasshopper/blob/master/node_modules/gh-core/lib/db.js#L187 for how we do that

simong:
Thanks! I've merged this into master

node_modules/gh-admins/lib/api.js
mrvisser: :+1:

PR #115, opened 2015-06-16
Create new groups for a `course`, `subject` or `part` when importing

simong:


node_modules/gh-orgunit/lib/internal/dao.js
```diff
@@ -982,8 +982,42 @@ var _updateOrCreateOrgUnit = function(appId, orgUnitToImport, parent, callback)
             'published': orgUnitToImport.published,
             'parentId': parent.id
         };
-        return createOrgUnit(appId, orgUnitToImport.displayName, orgUnitToImport.type, parent.GroupId, opts, callback);
+
+        // TODO: Refactor this check out of the DAO
+        _getOrCreateGroup(appId, orgUnitToImport.type, parent, function(err, groupId) {
```
mrvisser: Do we actually want this to be a "get or create", or should it already exist?

mrvisser: nm I get it. If it's a part, then we want it to have its own group created for it. If it's a module, then we want to use the parent course's group.

mrvisser: No, if it's a module, we want it to use the part's group. If it's a part, we want a new group created for it. part > module.

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -774,6 +774,71 @@ describe('Organisational units - import', function() {
...
+
+                            // Each subject should have a different group id
+                            assert.strictEqual(subjectGroupIds.length, 2);
+
+                            // Each part should have a different group id
+                            assert.strictEqual(partGroupIds.length, 4);
+
+                            // Parts should NOT inherit the subject or course's group id
+                            assert.strictEqual(_.intersection([course], subjectGroupIds).length, 0);
+                            assert.strictEqual(_.intersection([course], partGroupIds).length, 0);
```
mrvisser: It looks like we are intersecting an array with one course *object* with a list of string ids, that can't be right?

node_modules/gh-orgunit/tests/test-import.js
```diff
@@ -774,6 +774,71 @@ describe('Organisational units - import', function() {
...
+
+                            // Events should inherit the parts' group ids
+                            SeriesTestsUtil.getAllSeriesEvents(global.tests.admins.cam2013.client, _.pluck(series, 'id'), function(eventsPerSeries) {
+                                var eventGroupIds = _.chain(eventsPerSeries)
+                                    .values()
+                                    .flatten()
+                                    .pluck('GroupId')
+                                    .uniq()
+                                    .value()
+                                    .sort();
```
mrvisser: This formation looks like a gun

mrvisser:
Just one comment about I think a couple faulty assertions. Back to @simong and then this is good to be merged.

mrvisser:
:+1:

PR #116, opened 2015-06-17
Terms and Conditions

simong:
I cleaned this up and rebased it against master. This allows application/global administrators to configure a terms and conditions.

Note that this needs the following migrations:
```
alter table "Configs" add "termsAndConditionsText" text default ''::text;
alter table "Configs" add "termsAndConditionsEnabled" boolean default false;
alter table "Configs" add "termsAndConditionsLastUpdate" timestamp with time zone;
```

and requires this commit to retain a functioning UI:
https://github.com/fronteerio/grasshopper-ui/commit/f6142de484f1f223c0f0c3555000181582304d20

### DIFF
--- a/node_modules/gh-apps/lib/api.js
+++ b/node_modules/gh-apps/lib/api.js
@@ -324,7 +324,7 @@ var getTermsAndConditions = module.exports.getTermsAndConditions = function(ctx,
             return callback(err);
         }
 
-        var termsAndConditions = _getTermsAndConditions(app.id);
+        var termsAndConditions = getTermsAndConditionsForApp(app);
         return callback(null, termsAndConditions);
     });
 };
@@ -332,19 +332,17 @@ var getTermsAndConditions = module.exports.getTermsAndConditions = function(ctx,
 /**
  * Get the Terms and Conditions for an application.
  *
- * This method performs no validation on the given application id and should only be used when
- * you're sure the application exists. It's only exported because some dependencies shouldn't incur
- * the overhead of an extra DB query that happens through `getTermsAndConditions`
+ * This method performs no validation on the given application and should only be used when
+ * you're sure the application exist
  *
- * @param  {Number}         id                              The id of the application for which to retrieve the terms and conditions
+ * @param  {App}            app                             The application for which to retrieve the terms and conditions
  * @return {Object}         termsAndConditions              The Terms and Conditions for the application
  * @throws {Error}                                          Error when the application does not exist
- * @api private
  */
-var _getTermsAndConditions = module.exports._getTermsAndConditions = function(id) {
+var getTermsAndConditionsForApp = module.exports.getTermsAndConditionsForApp = function(app) {
     return {
-        'text': ConfigAPI.config(id).termsAndConditionsText,
-        'enabled': ConfigAPI.config(id).termsAndConditionsEnabled,
-        'lastUpdate': ConfigAPI.config(id).termsAndConditionsLastUpdate
+        'text': ConfigAPI.config(app.id).termsAndConditionsText,
+        'enabled': ConfigAPI.config(app.id).termsAndConditionsEnabled,
+        'lastUpdate': ConfigAPI.config(app.id).termsAndConditionsLastUpdate
     };
 };
--- a/node_modules/gh-apps/lib/api.js
+++ b/node_modules/gh-apps/lib/api.js
@@ -18,6 +18,7 @@
 
 var _ = require('lodash');
 
+var ConfigAPI = require('gh-config');
 var GrasshopperUtil = require('gh-core/lib/util');
 var log = require('gh-core/lib/logger').logger('gh-apps');
 var TenantsAPI = require('gh-tenants');
@@ -73,7 +74,7 @@ var getApp = module.exports.getApp = function(ctx, id, callback) {
 
     // Validation
     var validator = new Validator();
-    validator.check(id, {'code': 400, 'msg': 'A valid tenant id must be provided'}).isInt();
+    validator.check(id, {'code': 400, 'msg': 'A valid app id must be provided'}).isInt();
     if (validator.hasErrors()) {
         return callback(validator.getFirstError());
     }
@@ -293,4 +294,57 @@ var updateAppAdmins = module.exports.updateAppAdmins = function(ctx, id, adminUp
             AppsDAO.updateAppAdmins(adminUpdates, callback);
         });
     });
+};
+
+/* Terms and Conditions */
+
+/**
+ * Get the Terms and Conditions for an application
+ *
+ * @param  {Context}        ctx                             Standard context containing the current user and the current app
+ * @param  {Number}         id                              The id of the application for which to retrieve the terms and conditions
+ * @param  {Function}       callback                        Standard callback function
+ * @param  {Object}         callback.err                    An error object, if any
+ * @param  {Object}         callback.termsAndConditions     The Terms and Conditions for the application
+ */
+var getTermsAndConditions = module.exports.getTermsAndConditions = function(ctx, id, callback) {
+    // Ensure that the app id is a valid number
+    id = GrasshopperUtil.getNumberParam(id);
+
+    // Validation
+    var validator = new Validator();
+    validator.check(id, {'code': 400, 'msg': 'A valid app id must be provided'}).isInt();
+    if (validator.hasErrors()) {
+        return callback(validator.getFirstError());
+    }
+
+    // Verify that the provided application exists
+    getApp(ctx, id, function(err, app) {
+        if (err) {
+            return callback(err);
+        }
+
+        var termsAndConditions = _getTermsAndConditions(app.id);
+        return callback(null, termsAndConditions);
+    });
+};
+
+/**
+ * Get the Terms and Conditions for an application.
+ *
+ * This method performs no validation on the given application id and should only be used when
+ * you're sure the application exists. It's only exported because some dependencies shouldn't incur
+ * the overhead of an extra DB query that happens through `getTermsAndConditions`
+ *
+ * @param  {Number}         id                              The id of the application for which to retrieve the terms and conditions
+ * @return {Object}         termsAndConditions              The Terms and Conditions for the application
+ * @throws {Error}                                          Error when the application does not exist
+ * @api private
+ */
+var _getTermsAndConditions = module.exports._getTermsAndConditions = function(id) {
+    return {
+        'text': ConfigAPI.config(id).termsAndConditionsText,
+        'enabled': ConfigAPI.config(id).termsAndConditionsEnabled,
+        'lastUpdate': ConfigAPI.config(id).termsAndConditionsLastUpdate
+    };
 };
--- a/node_modules/gh-apps/lib/rest.js
+++ b/node_modules/gh-apps/lib/rest.js
@@ -184,5 +184,6 @@ var _getTermsAndConditions = function(req, res) {
         return res.status(200).send(termsAndConditions);
     });
 };
+
 GrassHopper.globalAdminRouter.on('get', '/api/apps/:id/termsAndConditions', _getTermsAndConditions);
 GrassHopper.appRouter.on('get', '/api/apps/:id/termsAndConditions', _getTermsAndConditions);
--- a/node_modules/gh-apps/lib/rest.js
+++ b/node_modules/gh-apps/lib/rest.js
@@ -163,3 +163,26 @@ var _updateAppAdmins = function(req, res) {
 
 GrassHopper.globalAdminRouter.on('post', '/api/apps/:id/admins', _updateAppAdmins);
 GrassHopper.appRouter.on('post', '/api/apps/:id/admins', _updateAppAdmins);
+
+/**
+ * @REST getTermsAndConditions
+ *
+ * Get the Terms and Conditions
+ *
+ * @Server      admin,app
+ * @Method      GET
+ * @Path        /apps/{id}/termsAndConditions
+ * @PathParam   {number}            id              The id of the app for which to get the Terms and Conditions
+ * @Return      {TermsAndConditions}                The Terms and Conditions for the app
+ */
+var _getTermsAndConditions = function(req, res) {
+    AppsAPI.getTermsAndConditions(req.ctx, req.params.id, function(err, termsAndConditions) {
+        if (err) {
+            return res.status(err.code).send(err.msg);
+        }
+
+        return res.status(200).send(termsAndConditions);
+    });
+};
+GrassHopper.globalAdminRouter.on('get', '/api/apps/:id/termsAndConditions', _getTermsAndConditions);
+GrassHopper.appRouter.on('get', '/api/apps/:id/termsAndConditions', _getTermsAndConditions);
--- a/node_modules/gh-core/lib/api.js
+++ b/node_modules/gh-core/lib/api.js
@@ -174,6 +174,19 @@ var _initialiseExpressServers = function(config, callback) {
     // Initialize the Passport authentication strategies
     AuthAPI.init(config);
 
+    /*!
+     * Add middleware that will check if the user has accepted the Terms and Conditions, if enabled.
+     * If the user hasn't accepted the Terms and Conditions, all POST requests (excluding whitelisted post requests) will be prevented.
+     */
+    module.exports.appServer.use(function(req, res, next) {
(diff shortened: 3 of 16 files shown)


simong:
Assigning to @mrvisser for review

node_modules/gh-core/lib/api.js
```diff
@@ -192,3 +205,14 @@ var _initialiseExpressServers = function(config, callback) {
...
 };
+
+/**
+ * Check if a URL is whitelisted from the Terms and Conditions requirements
+ *
+ * @param  {String}     url     The URL to check
+ * @return {Boolean}            `true` if the user doesn't have to accept the Terms and Conditions in order to POST to this url, `false` otherwise
+ * @api private
+ */
+var _isWhiteListed = function(url) {
```
mrvisser: Can you give a more meaningful name to this function? "whitelist" is a bit too generic. Maybe `_requiresTermsAndConditions`

node_modules/gh-core/lib/db.js
```diff
@@ -803,6 +806,9 @@ var _setUpModel = function(sequelize) {
      * @property {Boolean}      analyticsTrackingId                 The Segment.io analytics tracking identifier
      * @property {String}       academicYear                        The academic year for an application
      * @property {String}       statsd                              The ip, port and prefix that can be used to interact with statsd. Should be formatted as `ip:port:prefix`
+     * @property {String}       termsAndConditionsText              The terms and conditions for this page
```
mrvisser: `... for this application` ?

node_modules/gh-rest/lib/rest/user.js
```diff
@@ -98,7 +99,8 @@ module.exports = function(client) {
             'emailPreference': opts.emailPreference,
             'isAdmin': opts.isAdmin,
             'recaptchaChallenge': opts.recaptchaChallenge,
-            'recaptchaResponse': opts.recaptchaResponse
+            'recaptchaResponse': opts.recaptchaResponse,
+            'termsAndConditions': opts.termsAndConditions
```
mrvisser: Should this be a date of the last time the user has accepted terms and conditions? That way if they change, we can determine that users need to re-accept them.

mrvisser: I guess this would also imply that we need to always check the terms and conditions on every request. I haven't seen the whole PR yet, but do we do a query to invalidate all user's terms and conditions when the T&C are updated?

node_modules/gh-users/lib/api.js
```diff
@@ -784,3 +789,106 @@ var _getCalendarInfo = function(ctx, user, format) {
...
+ * @param  {Number}    appId        The application for which to retrieve the terms and conditions
+ * @return {Object}                 The Terms and Conditions for the application
+ */
+var getTermsAndConditions = module.exports.getTermsAndConditions = function(appId) {
+    return {
+        'text': ConfigAPI.config(appId).termsAndConditionsText,
+        'enabled': ConfigAPI.config(appId).termsAndConditionsEnabled,
+        'lastUpdate': ConfigAPI.config(appId).termsAndConditionsLastUpdate
+    };
+};
```
mrvisser: Is the user api really the best spot for this, as opposed to `gh-app` or `gh-config` ?

node_modules/gh-users/lib/api.js
```diff
@@ -784,3 +789,106 @@ var _getCalendarInfo = function(ctx, user, format) {
...
+        return false;
+
+    // Administrators don't need to accept anything
+    } else if (ctx.user.canAdmin(ctx.app.id)) {
+        return false;
+    }
+
+    // This application has Terms and Conditions. We need to check the user has accepted
+    // the Terms and Conditions since the last time they were updated
+    return (!ctx.user.termsAndConditions || ctx.user.termsAndConditions < termsAndConditions.lastUpdate);
```
mrvisser: I see we do check according to a date, so it may just be a documentation issue.

node_modules/gh-users/lib/rest.js
```diff
@@ -107,6 +107,7 @@ GrassHopper.appRouter.on('get', '/api/me', function(req, res) {
     if (req.ctx.user) {
         me = _.extend(me, {'anon': false}, req.ctx.user.toJSON());
     }
+    me = _.extend(me, {'termsAndConditions': UsersAPI.getTermsAndConditionsStatus(req.ctx)});
```
mrvisser: I don't think we need to assign this to `me` as it will update the first argument.

node_modules/gh-users/lib/rest.js
```diff
@@ -239,11 +240,12 @@ GrassHopper.appRouter.on('get', '/api/users/:id/upcoming', _getUserUpcoming);
...
  * @Method      GET
- * @Path        /users/termsAndConditions
+ * @Path        /termsAndConditions
  * @Return      {TermsAndConditions}                        The Terms and Conditions for the current app
  */
-GrassHopper.appRouter.on('get', '/api/users/termsAndConditions', function(req, res) {
-    res.sendStatus(501);
+GrassHopper.appRouter.on('get', '/api/termsAndConditions', function(req, res) {
+    var termsAndConditions = UsersAPI.getTermsAndConditions(req.ctx.app.id);
+    return res.status(200).send(termsAndConditions);
```
mrvisser: This API URL here indicates maybe the T&C accessing stuff should go in `gh-app`

node_modules/gh-users/tests/test-terms-and-conditions.js
```diff
@@ -0,0 +1,296 @@
...
+            });
+        });
+    };
+
+    describe('System interaction', function() {
+
+        /**
+         * Test that verifies that user need to accept the Terms and Conditions when creating an account
+         */
+        it('verify users need to accept the Terms and Conditions when creating an account', function(callback) {
```
mrvisser: Can admins create accounts on behalf of users? If so, they should be able to create a user account while deferring acceptance of the T&C to when the user attempts to use the system I would think. That should also be tested.

node_modules/gh-users/tests/test-terms-and-conditions.js
```diff
@@ -0,0 +1,296 @@
...
+        });
+
+        /**
+         * Test that verifies that users cannot interact with the system when a Terms and Conditions comes into effect
+         */
+        it('verify regular users need to accept the Terms and Conditions before they can interact with the system', function(callback) {
+            setup(function(tenant, app, globalAdminClient, simon) {
+
+                // When the user tries to do anything, they need to accept the Terms and Conditions
+                SeriesTestsUtil.assertCreateSerieFails(simon.client, 'Test serie', null, 419, function() {
```
mrvisser: `Test series`

node_modules/gh-users/tests/test-terms-and-conditions.js
```diff
@@ -0,0 +1,296 @@
...
+
+                        // Accept the Terms and Conditions
+                        UsersTestsUtil.assertAcceptTermsAndConditions(simon.client, simon.profile.id, function() {
+
+                            // Verify the me feed confirms that we don't need to accept the Terms and Conditions anymore
+                            UsersTestsUtil.assertGetMe(simon.client, function(me) {
+                                assert.strictEqual(me.termsAndConditions.needsToAccept, false);
+
+                                // Verify we can interact with the system
+                                SeriesTestsUtil.assertCreateSerie(simon.client, 'Test serie', null, function() {
```
mrvisser: `Test series`

node_modules/gh-users/tests/test-terms-and-conditions.js
```diff
@@ -0,0 +1,296 @@
...
+                        'termsAndConditionsEnabled': true,
+                        'termsAndConditionsText': 'Oh lawdy'
+                    };
+                    ConfigTestsUtil.assertUpdateConfig(globalAdminClient, app.id, configuration, function() {
+
+                        // Verify anonymous users do not have to accept them
+                        TestsUtil.getAnonymousAppUserClient(app, function(anonymousClient) {
+                            UsersTestsUtil.assertGetMe(anonymousClient, function(me) {
+                                assert.strictEqual(me.termsAndConditions.needsToAccept, false);
+                                return callback();
```
mrvisser: Just confirming that there are no POST requests in the system that an anonymous user should be able to submit that we can verify here?

node_modules/gh-users/tests/test-terms-and-conditions.js
```diff
@@ -0,0 +1,296 @@
...
+                // Get the Terms and Conditions
+                UsersTestsUtil.assertGetTermsAndConditions(simon.client, function(termsAndConditions) {
+
+                    // Update the Terms and Conditions
+                    var configuration = {'termsAndConditionsText': 'Oh heavens no'};
+                    ConfigTestsUtil.assertUpdateConfig(globalAdminClient, app.id, configuration, function() {
+
+                        // Get the updated Terms and Conditions
+                        UsersTestsUtil.assertGetTermsAndConditions(simon.client, function(updatedTermsAndConditions) {
+                            assert.ok(updatedTermsAndConditions.lastUpdate > termsAndConditions.lastUpdate);
```
mrvisser: Can we maybe toss a 100ms delay in here? *eye twitches*

node_modules/gh-users/tests/util.js
```diff
@@ -338,6 +341,63 @@ var assertGetUsersFails = module.exports.assertGetUsersFails = function(client,
...
+ * @param  {Function}           callback                        Standard callback function
+ * @param  {Object}             callback.termsAndConditions     The Terms and Conditions
+ * @throws {AssertionError}                                     Error thrown when an assertion failed
+ */
+var assertGetTermsAndConditions = module.exports.assertGetTermsAndConditions = function(client, callback) {
+    client.user.getTermsAndConditions(function(err, termsAndConditions) {
+        assert.ok(!err);
+        assert.ok(termsAndConditions);
+        assert.ok(termsAndConditions.text);
+        assert.ok(termsAndConditions.lastUpdate);
```
mrvisser: We can probably do more such as `_.isString(text)`, `_.isDate(lastUpdate)`, etc... maybe overkill to use the `generateTestFunction()` and extensions though?

node_modules/gh-users/tests/util.js
```diff
@@ -338,6 +341,63 @@ var assertGetUsersFails = module.exports.assertGetUsersFails = function(client,
...
+};
+
+/**
+ * Assert that a user cannot accept the Terms and Conditions
+ *
+ * @param  {RestClient}         client                          The REST client to make the request with
+ * @param  {Number}             id                              The id of the user for which to accept the Terms and Conditions
+ * @param  {Number}             code                            The expected HTTP error code
+ * @param  {Function}           callback                        Standard callback function
+ * @throws {AssertionError}                                     Error thrown when an assertion failed
```
mrvisser: The parameter description is really far away from the parameter names :/ But not a big deal

node_modules/gh-users/tests/util.js
```diff
@@ -338,6 +341,63 @@ var assertGetUsersFails = module.exports.assertGetUsersFails = function(client,
...
+ */
+var assertAcceptTermsAndConditions = module.exports.assertAcceptTermsAndConditions = function(client, id, callback) {
+    var start = moment();
+    client.user.acceptTermsAndConditions(id, function(err, termsAndConditionsStatus) {
+        assert.ok(!err);
+        assert.ok(termsAndConditionsStatus);
+        assert.ok(termsAndConditionsStatus.accepted);
+        var accepted = moment(termsAndConditionsStatus.accepted);
+        assert.ok(accepted.diff(start) > 0);
```
mrvisser: I think this is brittle. I've seen same-ms updates when I did oauth, unfortunately. I think for the sanity check, we can do `>=` instead of `>`, and then have a test that specifically ensures the timestamp updates with a delay of 100ms or something.

mrvisser: Oh, also maybe the `TestsUtil.assertTimestamp` might be useful

mrvisser:
@simong is there any place those migration SQL's should be placed in grasshopper for posterity?

mrvisser:
Just a few review comments, back to @simong

node_modules/gh-rest/lib/rest/user.js
simong: We persist the timestamp of when the T&C change in the config for each application (`config.termsAndConditionsLastUpdate`).
We persist the timestamp of when a user last accepted the T&C on their user profile (`user.termsAndConditions`).

On each request we'll check whether the user has accepted the T&C since they last were updated. There's no need to scan the user table and invalidate records

node_modules/gh-users/tests/util.js
simong: I started out with `generateTestFunction` but couldn't think of enough assertions to add to warrant it's use. Also, I found it a bit confusing to work with :(

node_modules/gh-users/tests/util.js
simong: Yeah, it's consistent with the other function's their JSDoc though :(

node_modules/gh-users/lib/rest.js
simong: I'm not sure I understand what you're saying here.

simong:
Pushed some follow-up, back to @mrvisser for review

node_modules/gh-apps/lib/api.js
```diff
@@ -294,3 +295,56 @@ var updateAppAdmins = module.exports.updateAppAdmins = function(ctx, id, adminUp
...
+ * This method performs no validation on the given application id and should only be used when
+ * you're sure the application exists. It's only exported because some dependencies shouldn't incur
+ * the overhead of an extra DB query that happens through `getTermsAndConditions`
+ *
+ * @param  {Number}         id                              The id of the application for which to retrieve the terms and conditions
+ * @return {Object}         termsAndConditions              The Terms and Conditions for the application
+ * @throws {Error}                                          Error when the application does not exist
+ * @api private
+ */
+var _getTermsAndConditions = module.exports._getTermsAndConditions = function(id) {
```
mrvisser: I'd prefer if this just took in an App instead of an ID. That way it's clear that the app has to exist.

node_modules/gh-users/lib/rest.js
mrvisser: I'm suggesting to change it to:

`_.extend(me, {'termsAndConditions': UsersAPI.getTermsAndConditionsStatus(req.ctx)});`

Rather than do the assign (`=`) to the `me` variable.

node_modules/gh-apps/lib/api.js
mrvisser: I should add that then we can make it a legit public method and not a `_`-prefixed that is exported.

node_modules/gh-rest/lib/rest/user.js
```diff
@@ -82,6 +82,7 @@ module.exports = function(client) {
      * @param  {Boolean}        [opts.isAdmin]                  Whether the user is an app administrator
      * @param  {String}         [opts.recaptchaChallenge]       The identifier for the recaptcha challenge. Only required when the current user is not an app administrator
      * @param  {String}         [opts.recaptchaResponse]        The recaptcha response entered for the presented challenge. Only required when the current user is not an app administrator
+     * @param  {Boolean}        [opts.termsAndConditions]       Whether the user accepts the Terms and Conditions
```
mrvisser: This should be `{Date}` I think. Maybe the language could say:

`When the user last accepted the terms and conditions, if at all`

node_modules/gh-apps/lib/rest.js
```diff
@@ -163,3 +163,26 @@ var _updateAppAdmins = function(req, res) {
...
+var _getTermsAndConditions = function(req, res) {
+    AppsAPI.getTermsAndConditions(req.ctx, req.params.id, function(err, termsAndConditions) {
+        if (err) {
+            return res.status(err.code).send(err.msg);
+        }
+
+        return res.status(200).send(termsAndConditions);
+    });
+};
+GrassHopper.globalAdminRouter.on('get', '/api/apps/:id/termsAndConditions', _getTermsAndConditions);
```
mrvisser: Needs new line before this. (reposing because other was a commit comment and not PR)

mrvisser:
Just a couple notes. Also, wanted to double-check about the SQL migration, maybe have an `etc/` directory to record those scripts that were needed?

mrvisser:
Review is super minor. Once those items are followed up on this can be merged.

simong:
Pushed follow-up

PR #117, opened 2015-06-24
Expose the `shibbolethId` to users from the same application

simong:


node_modules/gh-users/tests/test-users.js
```diff
@@ -278,20 +278,20 @@ describe('Users', function() {
 
                 // Simon can view his own information
                 UsersTestsUtil.assertGetUser(simon.client, simon.profile.id, function(user) {
-                    UsersTestsUtil.assertUser(user, {'expectedUser': simon.profile, 'canViewAllFields': true});
+                    UsersTestsUtil.assertUser(user, {'expectedUser': simon.profile, 'canViewPrivateFields': true});
 
                     // Nico can retrieve the profile but not the personal information
                     UsersTestsUtil.assertGetUser(nico.client, simon.profile.id, function(user) {
-                        UsersTestsUtil.assertUser(user, {'expectedUser': simon.profile, 'canViewAllFields': false});
+                        UsersTestsUtil.assertUser(user, {'expectedUser': simon.profile});
```
mrvisser: This test is ensuring that Nico can see all the private fields now, but the comment indicates that shouldn't be the case.

mrvisser: Nm, the private flag defaults to false but all the others default to true. That's a really tough API contract to follow :(

PR #118, opened 2015-06-29
Remove email matching from user search

simong:
This is a stop-gap measure to try and reduce the time a query takes. More investigation/potentially alternative solutions will follow soon.

### DIFF
--- a/node_modules/gh-users/lib/internal/dao.js
+++ b/node_modules/gh-users/lib/internal/dao.js
@@ -142,24 +142,32 @@ var searchUsers = module.exports.searchUsers = function(appId, query, limit, off
     limit = limit || 10;
     offset = offset || 0;
 
-    var sequelize = DB.getSequelize();
-
-    // Construct a query that selects a set of rows in the Users table ordered
-    // on the sum of the displayName and shibbolethId similarities. Although
-    // this isn't perfect, it gives a reasonable approximation of a typeahead search as
-    // each of the three columns their values are split up in possible trigrams and indexed
-    var sqlQuery = 'SELECT "Users".* ';
-    sqlQuery += 'FROM "Users" ';
-    sqlQuery += 'WHERE "AppId" = ? and ("displayName" <-> ? < 0.8 or "shibbolethId" <-> ? < 0.8) ';
-    sqlQuery += 'ORDER BY (("displayName" <-> ?) + ("shibbolethId" <-> ?)) asc ';
-    sqlQuery += 'LIMIT ? OFFSET ? ';
-
     var options = {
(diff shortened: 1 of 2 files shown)


node_modules/gh-users/lib/internal/dao.js
```diff
@@ -145,20 +145,17 @@ var searchUsers = module.exports.searchUsers = function(appId, query, limit, off
...
-    // value from a null value. Because of the index on each column, the query itself is
-    // reasonably quick and accurate
+    // each of the three columns their values are split up in possible trigrams and indexed
     var sqlQuery = 'SELECT "Users".* ';
     sqlQuery += 'FROM "Users" ';
-    sqlQuery += 'WHERE "AppId" = ? ';
-    sqlQuery += 'ORDER BY (similarity("displayName", ?) + similarity(coalesce("email", ?), ?) + similarity(coalesce("shibbolethId", ?), ?)) DESC, "displayName" ASC, "id" ASC ';
-    sqlQuery += 'LIMIT ? OFFSET ?' ;
-    var coalescedValue = '';
+    sqlQuery += 'WHERE "AppId" = ? and ("displayName" <-> ? < 0.8 or "shibbolethId" <-> ? < 0.8) ';
```
nicolaasmatthijs: I wouldn't be opposed to dropping the similarity search and just doing a more standard ilike match

mrvisser:
Where is the `<->` operator documented? Reading the trgm documentation, it claims to only optimize on a `%` operator:

http://www.postgresql.org/docs/9.0/static/pgtrgm.html

simong:
I changed the `searchUsers` logic so that it is doing:
* full-text search on `displayName` (e.g., `%Gaere%` or `%sg55`)
* prefix search on `shibbolethId` (e.g., `Gaere%` or `sg55`)

These queries run in half a ms.

simong:
I made a few typo's in my earlier comment. It should be:

* full-text search on `displayName` (e.g., `%Gaere%` or `%sg55%`)
* prefix search on `shibbolethId` (e.g., `Gaere%` or `sg55%`)

node_modules/gh-users/lib/internal/dao.js
```diff
@@ -142,27 +142,32 @@ var searchUsers = module.exports.searchUsers = function(appId, query, limit, off
...
-        'model': DB.User
+        'where': {
+            'AppId': appId,
+            '$or': [
+                {
+                    // Do a full text match on the display name. We can't do prefix matching here
+                    // as the display name column could contains strings such as `J. Donaldson` and
+                    // users might search for `Don`
+                    'displayName': {
+                        '$iLike': '%' + query + '%'
```
mrvisser: `util.format`

